PR #1, opened 2018-03-07
Remove unnecessary include EGL/eglmesaext.h

elima:
This header is not apparently necessary, and it is not present in non-mesa SDKs.


elima:
BTW, this include is preventing the backend from building on i.MX6 platforms with Vivante drivers.

PR #3, opened 2018-03-22
Add support for building using Meson

aperezdc:
I did this just as a proof of concept to see what it would take to migrate a CMake project to Meson, now that Buildroot has some support upstream for Meson — Feel free to skip merging this PR, or defer it for later, but personally I find Meson the less annoying of the existing configuration/build systems so my personal opinion is that this is a change for better :wink:

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -90,6 +90,7 @@ unstable_api_headers = [
 add_project_arguments(
 	'-DWPE_FDO_COMPILATION=1',
 	'-DG_LOG_DOMAIN="WPE-FDO"',
+	'-D_LARGEFILE64_SOURCE=1',
 	language: ['c', 'cpp']
 )
 
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,265 @@
+project('WPEBackend-fdo', ['c', 'cpp'],
+	meson_version: '>=0.49',
+	default_options: [
+		'b_ndebug=if-release',
+		'c_std=c99',
+		'cpp_eh=none',
+		'cpp_std=c++11',
+	],
+	license: 'BSD-2-Clause',
+	version: '1.7.0',
+)
+
+# This refers to the API level provided. This is modified only with major,
+# breaking changes, which need modifications in programs using the library
+# before they can be compiled again.
+api_version = '1.0'
+
+# Before making a release, the soversion array should be modified.
+# The string is of the form [C, R, A].
+# - If interfaces have been changed or added, but binary compatibility has
+#   been preserved, change to [C+1, 0, A+1].
+# - If binary compatibility has been broken (eg removed or changed interfaces)
+#   change to [C+1, 0, 0]
+# - If the interface is the same as the previous version, use [C, R+1, A].
+soversion = [7, 0, 6]
+
+# Split the *project* version into its components.
+version_info = meson.project_version().split('.')
+version_info = {
+	'PROJECT_VERSION_MAJOR': version_info[0],
+	'PROJECT_VERSION_MINOR': version_info[1],
+	'PROJECT_VERSION_PATCH': version_info[2],
+}
+
+# Mangle [C, R, A] into an actual usable *soversion*.
+soversion_major = soversion[0] - soversion[2]  # Current-Age
+soversion_minor = soversion[2]  # Age
+soversion_micro = soversion[1]  # Revision
+soversion = '@0@.@1@.@2@'.format(soversion_major, soversion_minor, soversion_micro)
+
+sources = [
+	'src/exported-buffer-shm.cpp',
+	'src/exported-image-egl.cpp',
+	'src/fdo.cpp',
+	'src/initialize-egl.cpp',
+	'src/initialize-shm.cpp',
+	'src/ipc.cpp',
+	'src/renderer-backend-egl.cpp',
+	'src/renderer-host.cpp',
+	'src/version.c',
+	'src/view-backend-exportable-fdo.cpp',
+	'src/view-backend-exportable-fdo-egl.cpp',
+	'src/view-backend-exportable-private.cpp',
+	'src/ws.cpp',
+	'src/ws-client.cpp',
+	'src/ws-egl.cpp',
+	'src/ws-shm.cpp',
+	'src/extensions/video-plane-display-dmabuf.cpp',
+	'src/extensions/video-plane-display-dmabuf-receiver.cpp',
+	'src/linux-dmabuf/linux-dmabuf.cpp',
+	'src/linux-dmabuf/linux-dmabuf-protocol.c',
+]
+
+api_headers = [
+	'include/wpe-fdo/exported-image-egl.h',
+	'include/wpe-fdo/exported-buffer-shm.h',
+	'include/wpe-fdo/initialize-egl.h',
+	'include/wpe-fdo/view-backend-exportable.h',
+	'include/wpe-fdo/view-backend-exportable-egl.h',
+	'include/wpe-fdo/fdo-egl.h',
+	'include/wpe-fdo/fdo.h',
+
+	# Generated API headers.
+	configure_file(
+		input: 'include/wpe-fdo/version.h.cmake',
+		output: 'version.h',
+		configuration: version_info,
+	),
+]
+
+extensions_api_headers = [
+	'include/wpe-fdo/extensions/video-plane-display-dmabuf.h',
+]
+
+unstable_api_headers = [
+	'include/wpe-fdo/unstable/fdo-shm.h',
+	'include/wpe-fdo/unstable/initialize-shm.h',
+]
+
+add_project_arguments(
+	'-DWPE_FDO_COMPILATION=1',
+	'-DG_LOG_DOMAIN="WPE-FDO"',
+	language: ['c', 'cpp']
+)
+
+wayland_scanner_dep = dependency('wayland-scanner', required: false, native: true)
+if wayland_scanner_dep.found()
+	wayland_scanner = find_program(
+		wayland_scanner_dep.get_pkgconfig_variable('wayland_scanner'),
+		native: true
+	)
+else
+	wayland_scanner = find_program('wayland-scanner', native: true)
+endif
+
+r = run_command(wayland_scanner, '--version')
+assert(r.returncode() == 0,
+	'''Cannot execute wayland-scanner:
+	''' + r.stderr())
+wayland_scanner_version = r.stderr().split()
+if not (wayland_scanner_version.length() == 2)
+	wayland_scanner_version = r.stdout().split()
+endif
+wayland_scanner_code = 'code'
+if wayland_scanner_version.length() == 2
+	wayland_scanner_version = wayland_scanner_version[1]
+	if wayland_scanner_version.version_compare('>=1.15')
+		wayland_scanner_code = 'private-code'
+	endif
+	message('wayland-scanner version: ' + wayland_scanner_version)
+else
+	wayland_scanner_version = ''
+	message('wayland-scanner version: unknown')
+endif
+message('Using "' + wayland_scanner_code + '" as parameter to wayland-scanner')
+
+# Wayland extension: WPE bridge.
+wpe_bridge_client_proto_header = custom_target(
+	'wpe-bridge-client-proto-header',
+	input: 'src/bridge/wpe-bridge.xml',
+	output: '@BASENAME@-client-protocol.h',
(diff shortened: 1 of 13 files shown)


elima:
Didn't check my self, but I remember @philn saying Yocto doesn't yet support meson. This could be a problem for a number of use cases we have.

mcatanzaro:
Some of WebKit's dependencies have themselves already switched to meson, e.g. glib-networking, and more are coming soon. So if you're using Yocto, it seems that implementing meson support would be an urgent priority.

aperezdc:
I agree with @mcatanzaro that Meson will be unavoidable at some point, and build systems are already starting to catch up adding support for it. That being said, I precisely wrote that it's fine to revisit this PR at a later time :wink:

philn:
Meson support is available in OE git master but for some projects we currently use Rocko, which doesn't have meson support...

philn:
Maybe we can revive this at some point? CC @nirbheek

aperezdc:
> Maybe we can revive this at some point? CC @nirbheek

I have just pushed a rehashed (and greatly improved) patch. The build mimics all features supported by the CMake buildsystem, except for the fact that we need https://github.com/mesonbuild/meson/issues/5800 fixed to be able to build the documentation with HotDoc.

The CI scripts need some changes to use Meson instead of CMake; I'll try to come up with something.

meson.build
```diff
@@ -0,0 +1,236 @@
...
+api_headers = [
+	'include/wpe-fdo/exported-image-egl.h',
+	'include/wpe-fdo/initialize-egl.h',
+	'include/wpe-fdo/view-backend-exportable.h',
+	'include/wpe-fdo/fdo-egl.h',
+	'include/wpe-fdo/fdo.h',
+
+	# Generated API headers.
+	configure_file(
+		input: 'include/wpe-fdo/version.h.cmake',
```
aperezdc: Ah, this could be renamed as `version.h.in` instead of using the `.cmake` suffix; I'll upload a new revision with that change applied.

nirbheek:
> except for the fact that we need [mesonbuild/meson#5800](https://github.com/mesonbuild/meson/issues/5800) fixed to be able to build the documentation with HotDoc.

CC: @MathieuDuponchelle

MathieuDuponchelle:
see https://github.com/mesonbuild/meson/pull/6163 :)

aperezdc:
The suggested workaround (adding a space to the `extra_c_flags` items) works, and HotDoc now gets correctly invoked. I still need to figure out how to pass a the include directories for the dependencies so HotDoc can successfully parse the API headers.

aperezdc:
> The suggested workaround (adding a space to the `extra_c_flags` items) works, and HotDoc now gets correctly invoked. I still need to figure out how to pass a the include directories for the dependencies so HotDoc can successfully parse the API headers.

My first attempt was to pass a `dependencies:` keyword argument to `hotdoc.generate_doc()`, but the generated JSON configuration file for HotDoc does _not_ contain the needed libwpe directories in `c_include_directories`:

```json
{
    "c_include_directories": [
        "include",
        "/home/aperez/devel/wpe/WPEBackend-fdo/b"
    ],
    ...
}
```

So this didn't work as I expected :disappointed:

thiblahute:
> `dependencies`



> > The suggested workaround (adding a space to the `extra_c_flags` items) works, and HotDoc now gets correctly invoked. I still need to figure out how to pass a the include directories for the dependencies so HotDoc can successfully parse the API headers.
> 
> My first attempt was to pass a `dependencies:` keyword argument to `hotdoc.generate_doc()`, but the generated JSON configuration file for HotDoc does _not_ contain the needed libwpe directories in `c_include_directories`:
> 
> ```json
> {
>     "c_include_directories": [
>         "include",
>         "/home/aperez/devel/wpe/WPEBackend-fdo/b"
>     ],
>     ...
> }
> ```
> 
> So this didn't work as I expected disappointed

What did you use as dependency? In GStreamer we use the result of `declare_dependency` and that works.

aperezdc:
I will be updating this in a way similar to https://github.com/WebPlatformForEmbedded/libwpe/pull/9 to make the CI use both build systems (Meson + CMake) and comparing the built results in order to avoid breaking one of them.

In the same vein as for `libwpe` we will be adding the Meson build support in WPEBackend-fdo 1.8, and keeping both build systems until 1.10, at which point we will zap the CMake build system. This gives plenty of time (one year) for packagers to update.

aperezdc:
The CI configuration is shamelessly copied from
https://github.com/WebPlatformForEmbedded/libwpe/pull/65 but I could change it back to use Travis if we would want that.

aperezdc:
This was rebased on top of `master` and updated to make the CI pass ensuring that both build systems generate a library with the same exported and undefined symbols. This should be ready to merge now.

Ping reviewers :ping_pong:

mcatanzaro:
>   * -DBUILD_DOCS=ON becomes -Dbuild-docs=trye
>   * -DEXPORTABLE_EGL=ON|OFF becomes -Dexportable-egl=true|false

meson style is -Dbuild_docs=true (underscore rather than hyphen) and -Dexportable_egl=enabled (feature rather than bool). The hyphens were an early mistake of early meson project, and such options are being renamed at great pain to the projects that originally offered them. Then bool options also predated the arrival of feature options.

meson.build
```diff
@@ -0,0 +1,269 @@
...
+	license: 'BSD-2-Clause',
+	version: '1.7.0',
+)
+
+# This refers to the API level provided. This is modified only with major,
+# breaking changes, which need modifications in programs using the library
+# before they can be compiled again.
+api_version = '1.0'
+
+# Before making a release, the LT_VERSION string should be modified.
```
mcatanzaro, line 18: But there is no LT_VERSION, is there? :) Replace the comment with this one:

https://gitlab.gnome.org/GNOME/gnome-desktop/-/blob/1eb198b9c5fc1e92cf5aa83cae2ab1a50762358f/meson.build#L7

meson.build
```diff
@@ -0,0 +1,269 @@
...
+
+# Split the *project* version into its components.
+version_info = meson.project_version().split('.')
+version_info = {
+	'PROJECT_VERSION_MAJOR': version_info[0],
+	'PROJECT_VERSION_MINOR': version_info[1],
+	'PROJECT_VERSION_PATCH': version_info[2],
+}
+
+# Mangle [C, R, A] into an actual usable *soversion*.
```
mcatanzaro, line 35: Then you won't need any of this!

meson.build
```diff
@@ -0,0 +1,269 @@
...
+sources = [
+    'src/exported-buffer-shm.cpp',
+	'src/exported-image-egl.cpp',
+	'src/fdo.cpp',
+	'src/initialize-egl.cpp',
+	'src/initialize-shm.cpp',
+	'src/ipc.cpp',
+	'src/renderer-backend-egl.cpp',
+	'src/renderer-host.cpp',
+    'src/version.c',
```
mcatanzaro, line 50: Indentation

meson.build
```diff
@@ -0,0 +1,269 @@
...
+	'src/ws-shm.cpp',
+	'src/extensions/video-plane-display-dmabuf.cpp',
+	'src/extensions/video-plane-display-dmabuf-receiver.cpp',
+	'src/linux-dmabuf/linux-dmabuf.cpp',
+	'src/linux-dmabuf/linux-dmabuf-protocol.c',
+]
+
+api_headers = [
+	'include/wpe-fdo/exported-image-egl.h',
+    'include/wpe-fdo/exported-buffer-shm.h',
```
mcatanzaro, line 65: Indentation

meson_options.txt
```diff
@@ -0,0 +1,8 @@
+option('exportable-egl',
+	type: 'boolean',
+	value: true,
+	description: 'Enable the exportable EGL interface')
```
zdobersek, line 4: I think this option should be abandoned. I wouldn't add it to Meson, unless we'd want it for compatibility with CMake. Wherever it is, it shouldn't do anything, and the EGL-based exportable should be built and installed unconditionally.

aperezdc: Done in the latest iteration of the patch set :ok_hand:

aperezdc:
> > * -DBUILD_DOCS=ON becomes -Dbuild-docs=trye
> > * -DEXPORTABLE_EGL=ON|OFF becomes -Dexportable-egl=true|false
> 
> meson style is -Dbuild_docs=true (underscore rather than hyphen) and -Dexportable_egl=enabled (feature rather than bool). The hyphens were an early mistake of early meson project, and such options are being renamed at great pain to the projects that originally offered them. Then bool options also predated the arrival of feature options.

Done, fixed in the latest iteration of the patch set.

aperezdc:
> What did you use as dependency? In GStreamer we use the result of `declare_dependency` and that works.

I tried adding the following, but didn't work:

```meson
libwpe_doc = hotdoc.generate_doc(meson.project_name(),
    # ...
    dependencies: declare_dependency(
        link_with: lib,
        dependencies: deps,
        include_directories: include_directories('include'),
    )
)
```

(The `lib` variable is the build target for the `libWPEBackend-fdo` shared library, and `deps` contains its list of dependencies.)

meson.build
aperezdc: Doing as suggested results in a different installed library name, and a different shared object version as well. We need to make the build with Meson generate exactly the same as for CMake, so for now we don't have any other option but keep calculating the `soversion` in the same way as done for the CMake build :shrug:

aperezdc:
This should be ready for another round of review :ping_pong:

aperezdc:
Question: Should I also add one more commit removing the Travis CI configuration in favor of the GitHub Actions one?

meson.build
mcatanzaro: Just manually set it to the exact same thing produced by the CMake build. Currently that's libWPEBackend-fdo-1.0.so.1.6.1, so you can start the version number at 1.6.1, and go from there. Much easier than CRA.

If you *really* want to use CRA, then at least remove the references to libtool. That wasn't even used by CMake. :P

aperezdc: Then we would have different numbers in `CMakeLists.txt` and `meson.build`, and I would rather keep them the same. Let's keep using *(C, R, A)* for now, and we can replace it later on when we remove the CMake build system in about a year :wink:

src/ws.h
```diff
@@ -41,6 +41,7 @@ struct ExportableClient {
     virtual void exportBufferResource(struct wl_resource*) = 0;
     virtual void exportLinuxDmabuf(const struct linux_dmabuf_buffer *dmabuf_buffer) = 0;
     virtual void exportShmBuffer(struct wl_resource*, struct wl_shm_buffer*) = 0;
+    virtual ~ExportableClient() = default;
```
zdobersek, line 44: Nit: the defaulted virtual constructor fits better at the beginning of this struct declaration.

src/ipc.h
```diff
@@ -33,6 +33,7 @@ namespace FdoIPC {
 class MessageReceiver {
 public:
     virtual void didReceiveMessage(uint32_t messageId, uint32_t messageBody) = 0;
+    virtual ~MessageReceiver() = default;
```
zdobersek, line 36: Nit: the defaulted virtual constructor fits better at the beginning of this struct declaration.

meson_options.txt
```diff
@@ -0,0 +1,4 @@
+option('build_docs',
+	type: 'boolean',
+	value: false,
+	description: 'Build reference documentation (needs HotDoc)')
```
zdobersek, line 4: There's an equivalent to the `EXPORTABLE_EGL` CMake option missing here. I mentioned somewhere before (maybe here, can't remember) that the option should be made no-op and the EGL-oriented exportable interfaces always built and installed.

If necessary for whatever aspect of portability and compliance, then a similar no-op `EXPORTABLE_EGL`-equivalent option should be added here. CMakeLists.txt should also be modified to build and install the EGL-oriented interfaces unconditionally.

src/ipc.h
aperezdc: Changed!

meson_options.txt
aperezdc: Instead of making it a no-op, I would rather remove the `option()` from `CMakeLists.txt`, because CMake will still accepts random undeclared `-DFOO=` options in the command line and will warn about them: existing build scripts won't break, but they will get a warning.

WDYT?

zdobersek: Sounds fine to me.

aperezdc:
I think this should now be ready for merging. I am leaving the removal of the Travis CI configuration for a separate PR to avoid piling up more things here :smile:

PR #10, opened 2018-04-25
Finishing touches

mcatanzaro:
* Rebased Zan's testing branch on master
 * Enforce single-header includes (and adapt for WPEBackend doing so)
 * Pad the sole public vtable
 * Add license headers
 * Remove CMake export
 * Add API version to installed files
 * Install files to proper GNU install directories


include/wpe-fdo/fdo-egl.h
```diff
@@ -0,0 +1,39 @@
...
+ * HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifdef __WEBKIT_WEB_EXTENSION_H__
```
zdobersek: Wrong macro?

include/wpe-fdo/fdo.h
```diff
@@ -0,0 +1,39 @@
...
+ * HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifdef __WEBKIT_WEB_EXTENSION_H__
```
zdobersek: Wrong macro?

include/wpe-fdo/fdo-egl.h
mcatanzaro: The goal there is to prevent these files from being included by mistake in a web extension. I presume that would always be a mistake, right?

zdobersek: Yes, it's just an odd thing to be preventing.

mcatanzaro: @zdobersek I can remove it if you like, but I can pretty much guarantee that confused bug reports will result, so I would recommend keeping it. With WebKitGTK+, we even get bug reports from people confused by the error message that the headers can't be included together. Having no error message will be worse. Anyway, whatever you prefer is fine with me.

mcatanzaro: Note also: https://github.com/WebPlatformForEmbedded/WPEBackend/commit/2bf3c3ae5f56d4d7383601f27c30f98cbe758a7d included the same checks

PR #14, opened 2018-05-23
wl improvements

zdobersek:
Move the Wayland connection handling into a separate thread.

Use a dedicated wl_event_queue for our operations on said Wayland connection.

Clean up all the Wayland resources in Backend destructor in renderer-backend-egl.

Flush the client after frame dispatch or buffer release commands are issued, immediately writing the messages to the display connection.


zdobersek:
Continued in #15.

PR #15, opened 2018-06-05
wl improvements

zdobersek:
Supersedes PR #14, approach there left us exposed to various threading issues.


elima:
I see a wayland proxy wrapper is being used. Can we avoid that?

src/renderer-backend-egl.cpp
```diff
@@ -96,139 +116,161 @@ class Backend {
...
+            wl_compositor_destroy(m_wl.compositor);
+        if (m_wl.registry)
+            wl_registry_destroy(m_wl.registry);
+        if (m_wl.eventQueue)
+            wl_event_queue_destroy(m_wl.eventQueue);
+        if (m_wl.displayWrapper)
+            wl_proxy_wrapper_destroy(m_wl.displayWrapper);
+
+        if (m_glib.socket)
+            g_object_unref(m_glib.socket);
```
aperezdc: I would just replace all the manual `NULL`-checking with `g_clear_pointer()` and `g_clear_object()`, e.g:

```c
g_clear_pointer (&m_wl.compositor, wl_compositor_destroy);
g_clear_pointer (&m_wl.registry, wl_registry_destroy);
// ...
g_clear_object (&m_glib.socket);
// ...
```

On top of making the code more concise, using these ensures that the pointers are always set to `NULL` which is handy to help find out use-after-free situations.

src/renderer-backend-egl.cpp
```diff
@@ -96,139 +116,161 @@ class Backend {
...
+        m_glib.frameSource = g_source_new(&s_sourceFuncs, sizeof(GSource));
+        g_source_set_priority(m_glib.frameSource, -70);
+        g_source_set_name(m_glib.frameSource, "WPEBackend-fdo::frame");
+        g_source_set_callback(m_glib.frameSource, [](gpointer userData) -> gboolean {
             auto* target = static_cast<Target*>(userData);
 
-            wl_callback_destroy(target->m_frameCallback);
-            target->m_frameCallback = nullptr;
+            wl_callback_destroy(target->m_wl.frameCallback);
+            target->m_wl.frameCallback = nullptr;
```
aperezdc: Ditto, `g_clear_pointer()` could be used here as well.

src/renderer-backend-egl.cpp
```diff
@@ -138,22 +138,15 @@ class Target {
...
+        g_clear_pointer(&m_wl.window, wl_egl_window_destroy);
+        g_clear_pointer(&m_wl.surface, wl_surface_destroy);
 
-        if (m_glib.socket)
-            g_object_unref(m_glib.socket);
+        g_clear_pointer(&m_wl.compositor, wl_compositor_destroy);
+        g_clear_pointer(&m_wl.registry, wl_registry_destroy);
+        g_clear_pointer(&m_wl.eventQueue, wl_event_queue_destroy);
+
+        g_clear_pointer(&m_glib.socket, g_object_unref);
```
elima: nitpick: for a GObject, I think [g_clear_object()](https://developer.gnome.org/gobject/unstable/gobject-The-Base-Object-Type.html#g-clear-object) is desirable.

PR #22, opened 2018-08-19
Use the libtool versioning scheme for libWPEBackend-fdo

aperezdc:
This is needed in preparation for the 1.0.0 release.

VersioningUtils.cmake is imported from the WebKit source tree:
https://trac.webkit.org/browser/webkit/trunk/Source/cmake/VersioningUtils.cmake

----

Related: https://github.com/WebPlatformForEmbedded/WPEBackend/pull/26

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,11 +1,23 @@
 cmake_minimum_required(VERSION 3.0)
 cmake_policy(VERSION 3.0)
 
-project(wpebackend-fdo VERSION 0.1)
+set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+include(VersioningUtils)
 
+SET_PROJECT_VERSION(1 0 0)
 set(WPEBACKEND_FDO_API_VERSION 0.1)
 
-set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+# Before making a release, the LT_VERSION string should be modified.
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,11 +1,15 @@
 cmake_minimum_required(VERSION 3.0)
 cmake_policy(VERSION 3.0)
 
-project(wpebackend-fdo VERSION 0.1)
+set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+include(VersioningUtils)
 
+SET_PROJECT_VERSION(1 0 0)
 set(WPEBACKEND_FDO_API_VERSION 0.1)
+CALCULATE_LIBRARY_VERSIONS_FROM_LIBTOOL_TRIPLE(LIBWPEBACKEND_FDO 1 0 0)
```
mcatanzaro: You should add the standard comment to improve the odds that it will be updated properly:

```
# Before making a release, the LT_VERSION string should be modified.
# The string is of the form C:R:A.
# - If interfaces have been changed or added, but binary compatibility has
#   been preserved, change to C+1:0:A+1
# - If binary compatibility has been broken (eg removed or changed interfaces)
#   change to C+1:0:0
# - If the interface is the same as the previous version, change to C:R+1:A
```

aperezdc: Done in the latest push :+1:

PR #23, opened 2018-08-20
Update to use libwpe 1.0.0

aperezdc:
The [new version of the dependency](https://wpewebkit.org/release/libwpe-1.0.0.html) renamed libWPEBackend to libwpe, and changed the API version to 0.2, so follow suit accordingly.


cmake/FindWPE.cmake
```diff
@@ -29,19 +29,19 @@
 # ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 find_package(PkgConfig)
-pkg_check_modules(PC_WPE QUIET wpe-0.1)
+pkg_check_modules(PC_WPE QUIET wpe-0.2)
```
mcatanzaro: Oh no, 0.2?

I guess it's too late now, but I would *definitely* have changed the API version to wpe-1.0.

Of course the API version is not necessarily expected to match the pkg-config version, but it looks much nicer to avoid 0.x version numbers now that we've gone stable.

We can either live with wpe-0.2, or else do a quick 1.0.1 release of libwpe to change the pkg-config version before anybody notices. Up to you. But if the goal is to not break the API from here on out, then wpe-0.2 seems like a quite strange API version to be stuck with.

aperezdc: Right, that *would* have been neat... A bit late now, so let's bump the API version to 1.0 next time there are changes.

PR #26, opened 2018-11-22
Return early if wpe_fdo_initialize_for_egl_display is called multiple…

carlosgcampos:
… times

Show an error message in case a different EGL display is provided.


aperezdc:
Merged to the 1.0 branch as 10ff8057613f8a3ef53a7e62e3886ad96d31d551.

PR #29, opened 2019-02-08
Add skeleton to build API documentation with HotDoc

aperezdc:
This adds support for building the documentation with a CMake option and the needed bits to make sure HotDoc will be invoked with the needed options. Actual documentation would be a follow-up.


CMakeLists.txt
```diff
@@ -108,3 +108,34 @@ install(
...
+        add_custom_target(Documentation ALL
+            ${HOTDOC} run
+                --project-name WPEBackend-fdo
+                --project-version ${PROJECT_VERSION}
+                --sitemap ${CMAKE_SOURCE_DIR}/docs/sitemap.txt
+                --output ${CMAKE_CURRENT_BINARY_DIR}/Documentation/
+                --c-smart-index
+                --extra-c-flags=-DWPE_FDO_COMPILATION
+                --c-sources
+                    ${CMAKE_SOURCE_DIR}/include/wpe-fdo/*.h
```
thiblahute: This is where the GtkDoc comment will be looked for, do you plan on documenting in the headers? (I prefer doing so personnaly, but I just want to make sure it is your intents here).

aperezdc: Personally I *do* like documentation in headers, because if I am missing something and I don't have the documentation at hand there's also the possibility of opening the installed `.h` files to take a look.

I am aware that most projects which follow GLib/GNOME conventions put the documentation comments in the `.c` files, so I am fine with that as well if other prefer it.

thiblahute: +2 for doc in headers then :-)

PR #32, opened 2019-02-22
Make wpe_fdo_initialize_for_egl_display return bool

carlosgcampos:


zdobersek:
@aperezdc any concerns regarding ABI change?

aperezdc:
@zdobersek As a matter of fact, I think right now it is a good moment to do such a change because the 2.24.x WPE WebKit releases are coming soon. As long as we coordinate to bump the API version here and apply the needed change to the WebKit repository, there should not be any problem.

Also, after merging this we may want to make a 1.1.90 release in preparation for the 1.2.x stable series. Right?

carlosgcampos:
Note this should be backwards comptaible, there's no need to recompile, existing applications would just ignore the return value.

aperezdc:
@carlosgcampos We also had some functions changed to receive `const` pointers (see #24 — binaries built against the old headers without `const` will still work, though).

philn:
Broke my build:
```
FAILED: CMakeFiles/WPEBackend-fdo.dir/src/ws.cpp.o 
/run/ccache/bin/c++   -DWAYLAND_1_10_OR_GREATER -DWPEBackend_fdo_EXPORTS -DWPE_FDO_COMPILATION -Iinclude -I. -Isrc -I/usr/include/glib-2.0 -I/usr/lib/glib-2.0/include -I/app/include/wpe-0.2 -O2 -g -std=c++11 -fno-exceptions -fno-rtti -fPIC -MD -MT CMakeFiles/WPEBackend-fdo.dir/src/ws.cpp.o -MF CMakeFiles/WPEBackend-fdo.dir/src/ws.cpp.o.d -o CMakeFiles/WPEBackend-fdo.dir/src/ws.cpp.o -c /run/build/wpebackend-fdo/src/ws.cpp
/run/build/wpebackend-fdo/src/ws.cpp: In function ‘bool WS::isEGLExtensionSupported(const char*, const char*)’:
/run/build/wpebackend-fdo/src/ws.cpp:323:40: error: ‘strlen’ was not declared in this scope
     int extensionLen = strlen(extension);
                                        ^
/run/build/wpebackend-fdo/src/ws.cpp:325:66: error: ‘strstr’ was not declared in this scope
     while ((extensionListPtr = strstr(extensionListPtr, extension))) {
                                                                  ^
```

aperezdc:
JFTR, the build issue was fixed by @philn in #35 (thanks!).

PR #36, opened 2019-02-26
Add API to query the exported EGL image size

carlosgcampos:


include/wpe-fdo/view-backend-exportable-egl.h
```diff
@@ -54,6 +55,9 @@ wpe_view_backend_exportable_fdo_egl_create(const struct wpe_view_backend_exporta
 void
 wpe_view_backend_exportable_fdo_egl_dispatch_release_image(struct wpe_view_backend_exportable_fdo* exportable, EGLImageKHR image);
 
+bool
+wpe_view_backend_exportable_fdo_egl_dispatch_query_image_size(struct wpe_view_backend_exportable_fdo* exportable, EGLImageKHR image, uint32_t* width, uint32_t* height);
```
aperezdc: I would just call this `wpe_view_backend_exportable_fdo_query_image_size()`. The “dispatch” word suggests that some kind of message is going to be sent somewhere and transported by some external entity (for example: snail-mail is *dispatched* by the post service)  but that is not the case here: the image is searched in the list of known buffers, and if found the dimensions returned (as far as I can understand).

aperezdc:
Also, quick question: do we want this merged to the 1.2.x release branch as well?

include/wpe-fdo/view-backend-exportable-egl.h
carlosgcampos: I agree, but I added the dispatch for consistency with existing API wpe_view_backend_exportable_fdo_egl_dispatch_release_image().

carlosgcampos:
> Also, quick question: do we want this merged to the 1.2.x release branch as well?

@zdobersek has plans to add an opaque struct wrapping the EGL image, so I'll wait for that new API, because in that case we will just add getters for width/height

include/wpe-fdo/view-backend-exportable-egl.h
aperezdc: Note how there is `wpe_view_backend_exportable_fdo_create()`, without “dispatch”, and the type returned is `struct wpe_view_backend_exportable_fdo`. 

Probably the `_egl_dispatch_release_image()` function has dispatch in the name because it ends up causing an IPC message (or so I understand, @zdobersek would know for sure).

carlosgcampos: dispatch_release_image is pretty much the same as query_image_size, it looks for the image in the list and releases it, there's no IPC involved at all.

elima: Maybe @zdobersek can chime in, but my understanding is that the "dispatch" keyword in the API is a convention to hint at the direction of the API: calls from the app into the backend use "dispatch", while calls from backend into app use the exportable interface.
I don't feel strongly about it, my gut feeling is that we could remove all those "disptach" from the API. But since they are there, I would rather be consistent.

carlosgcampos:
This functionality will part of patch in https://github.com/Igalia/WPEBackend-fdo/pull/37

PR #37, opened 2019-04-01
Add wpe_fdo_egl_exported_image struct to wrap the EGL image

carlosgcampos:
A new vfunc has been added to wpe_view_backend_exportable_fdo_egl_client
that receives a struct wpe_fdo_egl_exported_image. The image is
created associated to its buffer resource, and released when the buffer
is destroyed if the user released it back.

### DIFF
--- a/src/exported-image-egl.cpp
+++ b/src/exported-image-egl.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017, 2018 Igalia S.L.
+ * Copyright (C) 2019 Igalia S.L.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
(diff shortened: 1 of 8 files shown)


src/exported-image-egl.cpp
```diff
@@ -0,0 +1,63 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "view-backend-exportable-fdo-egl-private.h"
+#include "ws.h"
+#include <wpe-fdo/exported-image-egl.h>
```
zdobersek: <cassert> is needed for the assert() call bellow.

zdobersek: `<cassert>` inclusion ...

src/view-backend-exportable-fdo-egl.cpp
```diff
@@ -106,6 +106,90 @@ class ClientBundleEGL final : public ClientBundle {
     std::list<struct buffer_data*> m_buffers;
 };
 
+class ClientBundleEGL final : public ClientBundle {
```
zdobersek: I think this still needs a virtual destructor.

PR #39, opened 2019-05-15
Improve how CMake looks for wayland-scanner

aperezdc:
This improves `FindWaylandScanner.cmake` to allow it to find the `wayland-scanner` tool in any of the following ways (in order of preference):

1. By passing `-DWAYLAND_SCANNER=path/to/wayland-scanner` to CMake.
2. Looking for wayland-scanner in the current `$PATH`.
3. Figuring out the location from `pkg-config` (as before).

This will make the package friendlier to cross-compilation, and then build systems can either prepend their locations for host tools to `$PATH` (covered by 2. above), or they can specify the tool path in the command line (covered by 1. above) if they need to be more specific.

The motivation for this patch is to avoid kludges like the following:

  https://patchwork.ozlabs.org/comment/2172010/

### DIFF
--- a/cmake/FindWaylandScanner.cmake
+++ b/cmake/FindWaylandScanner.cmake
@@ -28,15 +28,90 @@
 # OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
 # ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-find_package(PkgConfig)
-pkg_check_modules(WAYLAND_SCANNER wayland-scanner)
+set(WAYLAND_SCANNER "" CACHE FILEPATH "Path to the wayland-scanner tool")
 
-pkg_get_variable(WAYLAND_SCANNER wayland-scanner wayland_scanner)
-if (WAYLAND_SCANNER_VERSION VERSION_LESS "1.15")
-    set(WAYLAND_SCANNER_CODE_ARG "code")
-else ()
-    set(WAYLAND_SCANNER_CODE_ARG "private-code")
+function(wayland_scanner_get_version _result _exepath)
+    set(_version "")
+    set(_status -1)
+    if (EXISTS "${_exepath}")
+        execute_process(
+            COMMAND "${_exepath}" "--version"
+            RESULT_VARIABLE _status
+            ERROR_VARIABLE _version
+            ERROR_STRIP_TRAILING_WHITESPACE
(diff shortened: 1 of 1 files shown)


cmake/FindWaylandScanner.cmake
```diff
@@ -28,15 +28,90 @@
...
+# Method 3: Try to find the executable using pkg-config, in case the
+#           tool was installed in a non-standard location.
+#
+if (NOT DEFINED WAYLAND_SCANNER OR NOT WAYLAND_SCANNER)
+    find_package(PkgConfig)
+    pkg_check_modules(WAYLAND_SCANNER_PC wayland-scanner)
+    if (WAYLAND_SCANNER_PC_FOUND)
+        pkg_get_variable(WAYLAND_SCANNER_PC_EXECUTABLE wayland-scanner wayland_scanner)
+        if (WAYLAND_SCANNER_PC_EXECUTABLE)
+            wayland_scanner_get_version(WAYLAND_SCANNER_VERSION "${WAYLAND_SCANNER_PC_EXECUTABLE}")
```
carlosgcampos: In the case of pkg-config we already have the version here, we don't need to execute the scanner to get its version.

aperezdc: I think it's better to try and run it to extract the version, to avoid the case in which the path returned by `pkg-config` points to a binary which cannot be executed on the host used for compilation… which would be useless, and was precisely what was happening when cross-building with Buildroot.

In this case I think it is better execute the program and to report failure (if it fails to run) when trying to  find an usable `wayland-scanner` at configuration time than to pretend things are fine and fail later during compilation in mysterious ways.

aperezdc:
I have double checked with Buildroot that having this patch applied indeed makes it pick the correct `wayland-scanner` from the host-tool directory:

```
···
-- Found WAYLAND_EGL: wayland-egl;wayland-client
-- Found WAYLAND_SCANNER: /home/aperez/buildroot/output-rpi3/host/bin/wayland-scanner
···
```

Note how the configuration log output copied above picked `…/host/bin/wayland-scanner`; without this patch it would have picked `…/target/usr/bin/wayland-scanner`, which in this particular test build is an ARM binary and thus not executable in the x86_64 build host.

Let's merge this :+1:

aperezdc:
I have also cherry picked this into the `wpebackend-fdo-1.2` branch as commit 9f77cf3c9fe6b543eb93fc1eab5d26cbd167c58f.

PR #40, opened 2019-05-30
Handle the case of new target created for the same view backend

carlosgcampos:
The client can remove a target and create a new one for the same view
backend. In that case we need to unregister the current view backend in
the wayland compositor before registering the one. This patch renames
WS::Instance::createSurface as registerSurface to clarify that it's not
creating a new surface but registering a new one for the given id. Also
ViewBackend::m_id has been renamed as m_surfaceId to clarify that's not
the backend identifier, but the current surface identifier.

### DIFF
--- a/src/renderer-backend-egl.cpp
+++ b/src/renderer-backend-egl.cpp
@@ -139,6 +139,11 @@ public:
 
     ~Target()
     {
+        if (m_wl.wpeBridgeId && m_glib.socket) {
+            uint32_t message[] = { 0x43, m_wl.wpeBridgeId };
+            g_socket_send(m_glib.socket, reinterpret_cast<gchar*>(message), 2 * sizeof(uint32_t), nullptr, nullptr);
+        }
+
         g_clear_pointer(&m_wl.frameCallback, wl_callback_destroy);
         g_clear_pointer(&m_wl.window, wl_egl_window_destroy);
         g_clear_pointer(&m_wl.surface, wl_surface_destroy);
@@ -223,6 +228,7 @@ public:
 
     void bridgeConnected(uint32_t bridgeID)
     {
+        m_wl.wpeBridgeId = bridgeID;
         uint32_t message[] = { 0x42, bridgeID };
         if (m_glib.socket)
             g_socket_send(m_glib.socket, reinterpret_cast<gchar*>(message), 2 * sizeof(uint32_t), nullptr, nullptr);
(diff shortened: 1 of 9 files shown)


carlosgcampos:
This is required by WebKitGTK, when leaving AC mode the renderer target is destroyed, and when entering again a new target is created.

src/renderer-backend-egl.cpp
```diff
@@ -139,6 +139,11 @@ class Target {
 
     ~Target()
     {
+        if (m_wl.wpeBridgeId && m_glib.socket) {
+            uint32_t message[] = { 0x43, m_wl.wpeBridgeId };
```
aperezdc: Ouch... using “magic” numbers does not make me super happy.

src/renderer-backend-egl.cpp
```diff
@@ -139,6 +139,11 @@ class Target {
...
+            g_socket_send(m_glib.socket, reinterpret_cast<gchar*>(message), 2 * sizeof(uint32_t), nullptr, nullptr);
```
aperezdc: How about using `sizeof (message)` instead of `2 * sizeof (uint32_t)` to let the compiler calculate the number of bytes to send here? Or even better: define a `MESSAGE_SIZE` constant.

src/view-backend-exportable-private.cpp
```diff
@@ -128,10 +146,17 @@ gboolean ViewBackend::s_socketCallback(GSocket* socket, GIOCondition condition,
...
-    if (len == sizeof(uint32_t) * 2 && message[0] == 0x42) {
-        auto& viewBackend = *static_cast<ViewBackend*>(data);
-        viewBackend.m_id = message[1];
-        viewBackend.m_client = WS::Instance::singleton().registerViewBackend(viewBackend.m_id, viewBackend);
+    if (len != sizeof(uint32_t) * 2)
+        return TRUE;
+
+    auto& viewBackend = *static_cast<ViewBackend*>(data);
+    switch (message[0]) {
+    case 0x42:
```
aperezdc: How about we define an `enum` for the message types? Something like:

```c
enum MessageType {
    RegisterSurface = 0x42,
    UnregisterSurface = 0x43,
};
```

and then in here:

```cpp
switch (message[0]) {
    case MessageType::RegisterSurface: /* ... */
    case MessageType::UnregisterSurface: /* ... */
    default: g_assert_not_reached();
}
```

src/renderer-backend-egl.cpp
carlosgcampos: I agree, existing code uses 0x42 for the other message so I just followed what we have.

src/view-backend-exportable-private.cpp
carlosgcampos: I agree too, and I thought about it, but again, I didn't want to change too much the existing code as part of this patch. I'm going to push a follow up patch to factor out the ipc code fixing all these style issues.

src/renderer-backend-egl.cpp
carlosgcampos: Same here, I'll fix it in a follow up patch on top of this one.

PR #41, opened 2019-06-07
Check pending wayland events in the queue on source preapre

carlosgcampos:
It's causing delays when there are multiple targets in the same process.

### DIFF
--- a/src/renderer-backend-egl.cpp
+++ b/src/renderer-backend-egl.cpp
@@ -44,7 +44,7 @@ struct Source {
     GPollFD pfd;
     struct wl_display* display;
     struct wl_event_queue* queue;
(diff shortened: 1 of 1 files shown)


carlosgcampos:
This is the reason for the poor rendering performance in WebKitGTK when having two related web views (sharing the same web process).

carlosgcampos:
hmm, sorry, I made a mistake in my tests, this patch doesn't fix the issue. I'll keep investigating.

carlosgcampos:
Sure, I'm working on a new patch, I'll reuse this PR

aperezdc:
It took me a bit of thinking to wrap my head about the changes, but they do look good. Thanks a lot for the patch!

PR #43, opened 2019-06-08
linux-dmabuf fixes

zdobersek:
1) compiler warning fix, changes the target type in the fourcc_mod_code macro to uint64_t. Replicates https://gitlab.freedesktop.org/mesa/drm/commit/b3c4c79e16f13a72e8124f69453a37135329f968.

2) One step to avoiding early releases of dmabuf-related wl_resources that can then cause tearing. If not nulled out here (like it's done for basic wl_resources), then the early release happens as soon as the next buffer is attached (see the attach implementation).


aperezdc:
Merged, thanks!

Issue #27, opened 2018-11-22
Add documentation for the public API


aperezdc:
Probably we would want to use [GTK-Doc](https://developer.gnome.org/gtk-doc-manual/stable/), which despite the name does not need GTK+ at all. That would allow cross-linking to the WPE WebKit API reference if needed.

See also https://github.com/WebPlatformForEmbedded/libwpe/issues/37 and [BWO#178900](https://bugs.webkit.org/show_bug.cgi?id=178900).

philn:
Or Hotdoc?

aperezdc:
@philn Is that [this HotDoc](https://hotdoc.github.io/)? I didn't know about it, looks like a good option as well!

philn:
Yes! GStreamer is migrating to it as well.

In-vincible:
@aperezdc is this issue up for grabs from beginners? If yes, please provide me with some more help about how to start with the work.

aperezdc:
@In-vincible Some help documenting the public API would definitely be appreciated. The tricky bit is to have the support for building the documentation; I have just posted #29 to wire HotDoc into the build system, once that patch merged, you should be able to do:

```sh
cmake -DBUILD_DOCS=ON
make Documentation
xdg-open Documentation/html/c-index.html 
```

Once that is working, you can start adding documentation next to function declarations using the [GTK-Doc syntax](https://developer.gnome.org/gtk-doc-manual/stable/documenting_syntax.html.en) (HotDoc uses the same syntax for C code), and re-run `make Documentation` to regenerate the HTML pages with your changes.

Please do ask if you need some more help from that point on, and remember that you can find us in the `#webkit` channel in the Freenode IRC :wave:

PR #45, opened 2019-06-10
wpe_fdo_egl_exported_image: add destroy-notify callback support.

zdobersek:
This should allow caching and subsequently safely destroying relevant
resources in user-facing applications.

### DIFF
--- a/src/exported-image-egl.cpp
+++ b/src/exported-image-egl.cpp
@@ -31,6 +31,14 @@
 extern "C" {
 
 __attribute__((visibility("default")))
+void
+wpe_fdo_egl_exported_image_set_destroy_notify(struct wpe_fdo_egl_exported_image* image, wpe_fdo_egl_exported_image_destroy_notify_t destroyNotify, void* destroyData)
+{
+    image->destroyNotify = destroyNotify;
+    image->destroyData = destroyData;
+}
+
+__attribute__((visibility("default")))
 uint32_t
 wpe_fdo_egl_exported_image_get_width(struct wpe_fdo_egl_exported_image* image)
 {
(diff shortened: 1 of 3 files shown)


src/exported-image-egl.cpp
```diff
@@ -30,6 +30,14 @@
 
 extern "C" {
 
+__attribute__((visibility("default")))
+void
+wpe_fdo_egl_exported_image_set_destroy_notify(struct wpe_fdo_egl_exported_image* image, wpe_fdo_egl_exported_image_destroy_notify_t destroyNotify, void* destroyData)
+{
+    image->destroyNotify = destroyNotify;
+    image->destroyData = destroyData;
```
aperezdc: Wouldn't it make sense that *if* a `destroyNotify` was already set previously, that the old callback is called before storing the new values? Not doing that could easily result in leaking the data passed as `destroyData`, if the callback is supposed to free it.

src/exported-image-egl.cpp
```diff
@@ -30,6 +30,14 @@
 
 extern "C" {
 
+__attribute__((visibility("default")))
+void
+wpe_fdo_egl_exported_image_set_destroy_notify(struct wpe_fdo_egl_exported_image* image, wpe_fdo_egl_exported_image_destroy_notify_t destroyNotify, void* destroyData)
```
aperezdc: This could use an API documentation comment. If we the missing ones little by little with each PR we would incrementally tackle #27 — I wouldn't block merging this PR due to this, though :wink:

src/exported-image-egl.cpp
zdobersek: Yes, that makes sense. I'll update it to do that.

aperezdc:
@zdobersek Do you have time to do the changes, or should we merge this as-is? I could do a follow-up PR myself with them myself.

aperezdc:
This has been stagnating for a long while so I am going to close it for now. Feel free to update and reopen again if/when this is relevant at some point.

PR #49, opened 2019-06-13
Remove the frame GLib source in render target

carlosgcampos:
We don't really need it since wayland events are already processed by
the wayland source in the target thread.


zdobersek:
I think you can also remove s_sourceFuncs? The `Target::s_sourceFuncs` ones.

carlosgcampos:
Indeed

PR #51, opened 2019-06-20
Rename the EGL members

psaavedra:
Renaming of the egl... members as s_ avoiding identifiers in our code
clashing with official names of functions is definitely a good idea to
avoid redeclaration errors like this:

```
  src/ws.cpp:74:33: error: 'void* (* eglCreateImageKHR)(EGLDisplay,
  EGLContext, EGLenum, EGLClientBuffer, const EGLint*)' redeclared as
  different kind of symbol
  static PFNEGLCREATEIMAGEKHRPROC eglCreateImageKHR;
```
The argument to eglGetProcAddress() has nothing to do with the name of
the variable. It is the name of a symbol from the EGL API:
<https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglGetProcAddress.xhtml>.

Also added non-null asserts to those pointers to make more evident
crashes due to null in the future (for example: changes in the name of
the symbols):

```
    assert(s_eglBindWaylandDisplayWL);
    assert(s_eglQueryWaylandBufferWL);
    assert(s_eglCreateImageKHR);
    assert(s_eglDestroyImageKHR);
    assert(s_eglQueryDmaBufFormatsEXT);
    assert(s_eglQueryDmaBufModifiersEXT);
```

### DIFF
--- a/src/ws.cpp
+++ b/src/ws.cpp
@@ -70,12 +70,12 @@ typedef EGLBoolean (* PFNEGLQUERYDMABUFMODIFIERSEXTPROC) (EGLDisplay dpy, EGLint
 #define EGL_DMA_BUF_PLANE3_MODIFIER_HI_EXT 0x344A
 #endif /* EGL_EXT_image_dma_buf_import_modifiers */
 
-static PFNEGLBINDWAYLANDDISPLAYWL eglBindWaylandDisplayWL;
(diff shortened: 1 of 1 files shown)


aperezdc:
Merged to the `wpebackend-fdo-1.2` branch as commit 318aaf08fd8e6923c90b5ce6a7bedeeb0ac97aab

PR #55, opened 2019-07-08
Support linux-dmabuf exports in wpe_view_backend_exportable_fdo_client

zdobersek:
Add the export_dmabuf_resource callback to the
wpe_view_backend_exportable_fdo_client interface. This callback provides an
wpe_view_backend_exportable_fdo_dmabuf_resource object that describes the
linux-dmabuf resource that's being exported.

The originating wl_resource is included in this new descriptor struct in order
to enable the application to release the resource through the existing
wpe_view_backend_exportable_fdo_dispatch_release_buffer() API. Other members of
the descriptor struct provide all the available information about the dmabuf
to enable reusing it through various other APIs like libgbm.

### DIFF
--- a/include/wpe-fdo/view-backend-exportable.h
+++ b/include/wpe-fdo/view-backend-exportable.h
@@ -39,9 +39,21 @@ extern "C" {
 struct wl_resource;
 struct wpe_view_backend_exportable_fdo;
 
+struct wpe_view_backend_exportable_fdo_dmabuf_resource {
+    struct wl_resource* buffer_resource;
+    uint32_t width;
+    uint32_t height;
+    uint32_t format;
+    uint8_t n_planes;
+    int32_t fds[4];
+    uint32_t strides[4];
+    uint32_t offsets[4];
+    uint64_t modifiers[4];
+};
+
 struct wpe_view_backend_exportable_fdo_client {
     void (*export_buffer_resource)(void* data, struct wl_resource* buffer_resource);
-    void (*_wpe_reserved0)(void);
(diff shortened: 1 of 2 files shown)


include/wpe-fdo/view-backend-exportable.h
```diff
@@ -39,9 +39,21 @@ extern "C" {
 struct wl_resource;
 struct wpe_view_backend_exportable_fdo;
 
+struct wpe_view_backend_exportable_fdo_dmabuf_resource {
+    struct wl_resource* buffer_resource;
+    uint32_t width;
+    uint32_t height;
+    uint32_t format;
+    uint8_t n_planes;
+    int32_t fds[4];
```
aperezdc: File descriptors should be of type `int`—which is the type used all around by the operating system for them. Is there some reason to use `int32_t` here instead?

zdobersek: It's what's used in the linux_dmabuf_attributes struct:
https://github.com/Igalia/WPEBackend-fdo/blob/master/src/linux-dmabuf/linux-dmabuf.h#L13

aperezdc: Fai enough :slightly_smiling_face: 

Let's merge this as-is for now, and I filed #56 as a cleanup to be done later on.

zdobersek: But Weston's implementation of linux_dmabuf_attributes is `int`, it was us who converted it to `int32_t` when the linux-dmabuf code was adopted.

PR #58, opened 2019-07-25
Arrange passing -Wall, check whether used compiler flags are supported

aperezdc:
The first commit arranges passing `-Wall` to the compiler, checking whether compiler flags are supported before adding them to the set of used ones, and using CMake's built-in C/C++ standard selection smarts. The second adds `final` to declarations to avoid a compiler warning (and it is good practice, too).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,10 +18,27 @@ CALCULATE_LIBRARY_VERSIONS_FROM_LIBTOOL_TRIPLE(LIBWPEBACKEND_FDO 4 0 3)
 
 project(wpebackend-fdo VERSION "${PROJECT_VERSION}")
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -fno-exceptions -fno-rtti")
-
+include(CheckCCompilerFlag)
+include(CheckCXXCompilerFlag)
 include(DistTargets)
 include(GNUInstallDirs)
+
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_C_STANDARD 99)
+
+foreach (cxxflag -fno-exceptions -fno-rtti -Wall)
+    check_cxx_compiler_flag("${cxxflag}" CXX_HAS_FLAG_${cxxflag})
+    if (CXX_HAS_FLAG_${cxxflag})
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${cxxflag}")
+    endif ()
+endforeach ()
+
+foreach (cflag -Wall)
+    check_c_compiler_flag("${cflag}" CC_HAS_FLAG_${cflag})
+    if (CC_HAS_FLAG_${cflag})
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${cflag}")
+    endif ()
+endforeach ()
 
 find_package(EGL REQUIRED)
 find_package(GLIB REQUIRED COMPONENTS gio)
--- a/src/view-backend-exportable-private.h
+++ b/src/view-backend-exportable-private.h
@@ -54,7 +54,7 @@ public:
     uint32_t initialHeight;
 };
 
-class ViewBackend : public WS::ExportableClient, public FdoIPC::MessageReceiver {
+class ViewBackend final : public WS::ExportableClient, public FdoIPC::MessageReceiver {
 public:
     ViewBackend(ClientBundle* clientBundle, struct wpe_view_backend* backend);
     ~ViewBackend();


carlosgcampos:
Does the compiler complain for not adding final to structs? it looks weird to me. I'm fine adding final for final classes that derive for another one, but I don't see how it helps for classes and structs with no parent.

aperezdc:
> Does the compiler complain for not adding final to structs? it looks weird to me. 

No, it doesn't… it just seemed a good idea to add `final` there because C++ allows “subclassing” a `struct` and also adding virtual methods to them. The idea is to avoid footguns: if someone subclasses a type, the compiler will bail because of the `final` and make us all think twice about how to do the subclassing properly.

> I'm fine adding final for final classes that derive for another one, but I don't see how it helps for classes and structs with no parent.

For the ones with no parent, it makes the compiler forbid subclassing them.

aperezdc:
Updated to add `final` only to the `ViewBackend` class.

@carlosgcampos, @mcatanzaro: WDYT about merging this now?

CMakeLists.txt
```diff
@@ -18,11 +18,28 @@ CALCULATE_LIBRARY_VERSIONS_FROM_LIBTOOL_TRIPLE(LIBWPEBACKEND_FDO 4 0 3)
...
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -fno-exceptions -fno-rtti")
-
+include(CheckCCompilerFlag)
+include(CheckCXXCompilerFlag)
 include(DistTargets)
 include(GNUInstallDirs)
 
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_C_STANDARD 99)
```
mcatanzaro: Nowadays GCC defaults to C++14 and C17. I guess you're aware that you're regressing the language versions here.

src/view-backend-exportable-private.h
```diff
@@ -54,7 +54,7 @@ class ClientBundle {
     uint32_t initialHeight;
 };
 
-class ViewBackend : public WS::ExportableClient, public FdoIPC::MessageReceiver {
+class ViewBackend final : public WS::ExportableClient, public FdoIPC::MessageReceiver {
```
mcatanzaro: Oooh, multiple inheritance! WS::ExportableClient and FdoIPC::MessageReceiver still both need virtual destructors.

aperezdc: Not really. A method (or destructor) needs to be marked `virtual` to make sure that the compiler ends up calling the one at the end of the inheritance chain when applying the `delete` operator to a pointer typed with one of the superclasses.

In the case of `ViewBackend`, the deletion is done in this way:

```cpp
auto* backend = reinterpret_cast<ViewBackend*>(data);
delete backend;
```

If the compiler knows that `ViewBackend` will not have subclasses (e.g. by marking ìt `final`) then it knows that it always needs to use `ViewBackend::~ViewBackend()` for destruction, and that this is always safe because there cannot be a more concrete implementation.

aperezdc: Side note: If the destructor was marked `virtual` and at the same time the `ViewBackend` class is marked `final`, then the compiler would [devirtualize](https://marcofoco.com/the-power-of-devirtualization/) the call—which effectively ends up choosing which version of the method/destructor will be used statically at compile time (thus skipping the dynamic vtable lookup at runtime).

CMakeLists.txt
aperezdc: Note that for C++ before this patch `-std=c++11` was being passed to the compiler already:

```cmake
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -fno-exceptions -fno-rtti")
```

On the C side, the only plain C code is `src/linux-dmabuf/linux-dmabuf-protocol.c` and the generated Wayland protocol code, none of which use C11 features, and that is the reason for choosing C99.

src/view-backend-exportable-private.h
mcatanzaro: > Here the compiler knows that we are deleting a `ViewBackend` (see the cast!), which will not ever have subclasses (when marked `final`). Therefore it knows that it always needs to use `ViewBackend::~ViewBackend()` for destruction, and that this is always safe because there cannot be a more concrete implementation.

I understand that, but if you ever change the way this ViewBackend instance is deleted, or if you ever create a different ViewBackend instance elsewhere and delete it from a parent class pointer, you leak the ViewBackend portion of the class, akin to slicing. This is much more of a footgun than leaving "final" off a struct declaration.

As a rule, any class that is derived from should have a virtual destructor. This is accepted C++ best-practice for decades now and is also the rule we use in WebKit.

aperezdc: There's only two ways in the way that the `ViewBackend` can be destroyed:

- A subclass is added, deletion still done through a `ViewBackend*` pointer: the compiler will tell you to remove final, and then warn about the class needing a virtual destructor. No problem.
- No subclass is added, but the destruction is done through a superclass pointer (`WS::ExportableClient*` or `FdoIPC::MessageReceiver`): the compiler will tell you that you need a virtual destructor because the type is not a final class. Again: no problem.

So there's no footgun now that we have compiler warnings enabled. How is the so-called “accepted best practice” not just cargo-culting in this case?

P.S: There is not such thing as “leaking” a slice of an object itself as a result from the wrong constructor being chosen, what happens is that some constructors are just not executed—if the code that does not run happens to dispose of resources those won't be properly disposed (of course: not cool), but the whole zone of memory allocated for *the instance itself* is always freed in full.

carlosgcampos:
> Updated to add `final` only to the `ViewBackend` class.
> 
> @carlosgcampos, @mcatanzaro: WDYT about merging this now?

LGTM

PR #60, opened 2019-07-26
Fix -Wdelete-non-virtual-dtor warning

mcatanzaro:
This was detected because a downstream manually adds -Wall to its build
flags. WPEBackend-fdo should consider turning on at least -Wall.

Note there is no functional bug here because ViewBackend is the
most-derived class being deleted, but this is certainly fragile.

### DIFF
--- a/src/view-backend-exportable-private.h
+++ b/src/view-backend-exportable-private.h
@@ -57,7 +57,7 @@ public:
 class ViewBackend : public WS::ExportableClient, public FdoIPC::MessageReceiver {
 public:
     ViewBackend(ClientBundle* clientBundle, struct wpe_view_backend* backend);
-    ~ViewBackend();
+    virtual ~ViewBackend();
 
     void initialize();
     int clientFd();


mcatanzaro:
Well I suppose #58 is a better solution than this.

aperezdc:
> Well I suppose #58 is a better solution than this.

Yes, there's no subclasses of the `ViewBackend` class, so it's better to mark it `final` and the compiler knows that it is fine to have a non-virtual destructor. I suppose we can close this in favor of the other PR, because I would rather have the code building with `-Wall` :wink:

mcatanzaro:
BTW the parents of `ViewBackend` really should have virtual destructors regardless, and that isn't fixed in your commit!

aperezdc:
> BTW the parents of `ViewBackend` really should have virtual destructors regardless, and that isn't fixed in your commit!

Not really, I [commented on the PR about that](https://github.com/Igalia/WPEBackend-fdo/pull/58#discussion_r307957824).

Issue #56, opened 2019-07-09
Use “int” for file descritors everywhere


aperezdc:
For example in the review of PR #55 we found out that there are some places where `int32_t` is being used for file descriptors, instead of `int` (which is the appropriate type).

zdobersek:
We adopted int32_t (at least in linux_dmabuf_attributes) on our own:
https://github.com/Igalia/WPEBackend-fdo/pull/55#discussion_r301436658
https://gitlab.freedesktop.org/wayland/weston/blob/master/libweston/linux-dmabuf.h#L43

aperezdc:
This has been fixed by PR #64, closing.

PR #64, opened 2019-07-28
Use int as the file descriptor type

zdobersek:
Convert back from int32_t to simply int when storing file descriptor values,
conforming to the usual Linux standards for this topic.


zdobersek:
This would address issue #56.

PR #65, opened 2019-07-28
Move Wayland client logic into WS::BaseBackend, WS::BaseTarget classes

zdobersek:
-

### DIFF
--- a/src/ws-client.cpp
+++ b/src/ws-client.cpp
@@ -1,0 +1,244 @@
+/*
+ * Copyright (C) 2019 Igalia S.L.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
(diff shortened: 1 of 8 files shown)


src/ws-client.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+    wpe_bridge_add_listener(m_wl.wpeBridge, &s_bridgeListener, this);
+    wpe_bridge_connect(m_wl.wpeBridge, m_wl.surface);
+    wl_display_roundtrip_queue(display, m_wl.eventQueue);
+}
+
+void BaseTarget::requestFrame()
+{
+    if (m_wl.frameCallback)
+        std::abort();
```
aperezdc: In a similar vein, `g_error()` would be nicer as it would also print a message, but this PR is only moving code around, so let's use #68 to track doing this later on.

PR #71, opened 2019-07-31
Fix syntax error FindEGL.cmake. Wrong uncommented lines.

psaavedra:


aperezdc:
Ouch, this shouldn't have sneaked in, thanks for the fix @psaavedra!

PR #73, opened 2019-09-07
wl_resource management fixes

zdobersek:
This tries to do a better job of tracking wl_resource lifetime and managing the resource release process in cases where the wl_resource was already destroyed.

### DIFF
--- a/src/exported-image-egl.cpp
+++ b/src/exported-image-egl.cpp
@@ -52,13 +52,3 @@ wpe_fdo_egl_exported_image_get_egl_image(struct wpe_fdo_egl_exported_image* imag
 }
 
 }
-
-void
-wpe_fdo_egl_exported_image_destroy(struct wpe_fdo_egl_exported_image* image)
-{
-    assert(image->eglImage);
-    WS::Instance::singleton().destroyImage(image->eglImage);
-    wl_list_remove(&image->bufferDestroyListener.link);
-
-    delete image;
-}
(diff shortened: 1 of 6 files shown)


src/exported-image-egl.cpp
```diff
@@ -52,13 +52,3 @@ wpe_fdo_egl_exported_image_get_egl_image(struct wpe_fdo_egl_exported_image* imag
 }
 
 }
-
-void
-wpe_fdo_egl_exported_image_destroy(struct wpe_fdo_egl_exported_image* image)
-{
-    assert(image->eglImage);
-    WS::Instance::singleton().destroyImage(image->eglImage);
-    wl_list_remove(&image->bufferDestroyListener.link);
```
carlosgcampos: This is no longer needed?

carlosgcampos:
LGTM

src/exported-image-egl.cpp
zdobersek: It's moved into the deleteImage() method on the ClientBundleEGL class. So everything still gets destroyed, just the code was localized to that class.

carlosgcampos: I meant the wl_list_remove specifically, github doesn't seem allow to select the lines in the review.

zdobersek: I think it's not needed. At this point the resource is already destroyed or is being destroyed, so removing that listener doesn't do much.

aperezdc:
Cherry picked as commit 5b8c3b8006cd7cbbb655da5bd322f91cb5f9788a in the `wpebackend-fdo-1.4` branch.

aperezdc:
Trying to cherry-pick this to the `wpebackend-1.2` branch results in quite a few conflicts.

@zdobersek Would you be so kind as to make a PR targeting the branch, if possible? Thanks!

PR #86, opened 2019-12-13
DMABuf video plane display protocol

philn:
Introducing a new protocol enabling client-side video rendering. In WebKit
decoded video frames will be optionally exported to DMABuf fds and passed on
through the IPC wire using this protocol. On the other end, the browser (Cog for
instance) should receive all the informations needed so it can position and
render the video surfaces directly.

With contributions from Žan Doberšek <zdobersek@igalia.com>.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,8 +76,13 @@ set(WPEBACKEND_FDO_PUBLIC_HEADERS
     include/wpe-fdo/fdo.h
 )
 
+set(WPEBACKEND_FDO_EXTENSIONS_PUBLIC_HEADERS
+    include/wpe-fdo/extensions/video-plane-display-dmabuf.h
+)
+
 set(WPEBACKEND_FDO_GENERATED_SOURCES
     ${CMAKE_BINARY_DIR}/bridge/wpe-bridge-protocol.c
+    ${CMAKE_BINARY_DIR}/video-plane-display-dmabuf/wpe-video-plane-display-dmabuf-protocol.c
 )
 
 set(WPEBACKEND_FDO_SOURCES
@@ -95,6 +100,9 @@ set(WPEBACKEND_FDO_SOURCES
     src/ws.cpp
     src/ws-client.cpp
 
+    src/extensions/video-plane-display-dmabuf.cpp
+    src/extensions/video-plane-display-dmabuf-receiver.cpp
+
     src/linux-dmabuf/linux-dmabuf.cpp
     src/linux-dmabuf/linux-dmabuf-protocol.c
 )
@@ -108,6 +116,18 @@ add_custom_command(
     COMMAND ${WAYLAND_SCANNER} ${WAYLAND_SCANNER_CODE_ARG} < ${CMAKE_SOURCE_DIR}/src/bridge/wpe-bridge.xml > ${CMAKE_BINARY_DIR}/bridge/wpe-bridge-protocol.c
     COMMAND ${WAYLAND_SCANNER} client-header < ${CMAKE_SOURCE_DIR}/src/bridge/wpe-bridge.xml > ${CMAKE_BINARY_DIR}/bridge/wpe-bridge-client-protocol.h
     COMMAND ${WAYLAND_SCANNER} server-header < ${CMAKE_SOURCE_DIR}/src/bridge/wpe-bridge.xml > ${CMAKE_BINARY_DIR}/bridge/wpe-bridge-server-protocol.h
+    VERBATIM
+)
+
+file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/video-plane-display-dmabuf)
+add_custom_command(
+    OUTPUT ${CMAKE_BINARY_DIR}/video-plane-display-dmabuf/wpe-video-plane-display-dmabuf-protocol.c
+    OUTPUT ${CMAKE_BINARY_DIR}/video-plane-display-dmabuf/wpe-video-plane-display-dmabuf-client-protocol.h
+    OUTPUT ${CMAKE_BINARY_DIR}/video-plane-display-dmabuf/wpe-video-plane-display-dmabuf-server-protocol.h
+    DEPENDS ${CMAKE_SOURCE_DIR}/src/extensions/wpe-video-plane-display-dmabuf.xml
+    COMMAND ${WAYLAND_SCANNER} ${WAYLAND_SCANNER_CODE_ARG} < ${CMAKE_SOURCE_DIR}/src/extensions/wpe-video-plane-display-dmabuf.xml > ${CMAKE_BINARY_DIR}/video-plane-display-dmabuf/wpe-video-plane-display-dmabuf-protocol.c
+    COMMAND ${WAYLAND_SCANNER} client-header < ${CMAKE_SOURCE_DIR}/src/extensions/wpe-video-plane-display-dmabuf.xml > ${CMAKE_BINARY_DIR}/video-plane-display-dmabuf/wpe-video-plane-display-dmabuf-client-protocol.h
+    COMMAND ${WAYLAND_SCANNER} server-header < ${CMAKE_SOURCE_DIR}/src/extensions/wpe-video-plane-display-dmabuf.xml > ${CMAKE_BINARY_DIR}/video-plane-display-dmabuf/wpe-video-plane-display-dmabuf-server-protocol.h
     VERBATIM
 )
 
@@ -136,6 +156,10 @@ install(
 install(
     FILES ${WPEBACKEND_FDO_PUBLIC_HEADERS}
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/wpe-fdo-${WPEBACKEND_FDO_API_VERSION}/wpe
+)
+install(
+    FILES ${WPEBACKEND_FDO_EXTENSIONS_PUBLIC_HEADERS}
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/wpe-fdo-${WPEBACKEND_FDO_API_VERSION}/wpe/extensions
 )
 
 configure_file(wpebackend-fdo.pc.in wpebackend-fdo-${WPEBACKEND_FDO_API_VERSION}.pc @ONLY)
@@ -166,6 +190,7 @@ if (BUILD_DOCS)
                 --extra-c-flags=-DWPE_FDO_COMPILATION
                 --c-sources
                     ${CMAKE_SOURCE_DIR}/include/wpe-fdo/*.h
+                    ${CMAKE_SOURCE_DIR}/include/wpe-fdo/extensions/*.h
                 --c-include-directories
                     ${CMAKE_SOURCE_DIR}/include
                     ${WPEBACKEND_FDO_INCLUDE_DIRECTORIES}
--- a/include/wpe-fdo/extensions/video-plane-display-dmabuf.h
+++ b/include/wpe-fdo/extensions/video-plane-display-dmabuf.h
@@ -1,0 +1,97 @@
+/*
+ * Copyright (C) 2019 Igalia S.L.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+ * PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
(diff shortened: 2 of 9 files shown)


CMakeLists.txt
```diff
@@ -76,8 +76,14 @@ set(WPEBACKEND_FDO_PUBLIC_HEADERS
     include/wpe-fdo/fdo.h
 )
 
+set(WPEBACKEND_FDO_EXTENSIONS_PUBLIC_HEADERS
+    include/wpe-fdo/extensions/video-plane-display-dmabuf.h
+    include/wpe-fdo/extensions/video-plane-display-dmabuf-receiver.h
```
carlosgcampos: We are forcing single header inclusion in fdo. I'm not sure if extensions are different or optional to require its own header, or one header per extension.

include/wpe-fdo/extensions/video-plane-display-dmabuf-receiver.h
```diff
@@ -0,0 +1,53 @@
...
+ * HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
```
carlosgcampos: Use #ifndef pattern in public headers.

include/wpe-fdo/extensions/video-plane-display-dmabuf-receiver.h
```diff
@@ -0,0 +1,53 @@
...
+
+#ifndef __video_plane_display_dmabuf_receiver_h__
```
carlosgcampos: It's here, so just remove the pragma once

include/wpe-fdo/extensions/video-plane-display-dmabuf-receiver.h
```diff
@@ -0,0 +1,53 @@
...
+ * PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
```
carlosgcampos: I don't think we should allow to include this header directly

include/wpe-fdo/extensions/video-plane-display-dmabuf-receiver.h
```diff
@@ -0,0 +1,53 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct wpe_video_plane_display_dmabuf_export;
+
+struct wpe_video_plane_display_dmabuf_receiver {
+    void (*handle_dmabuf)(void* data, struct wpe_video_plane_display_dmabuf_export*, uint32_t id, int fd, int32_t x, int32_t y, int32_t width, int32_t height, uint32_t stride);
+};
```
carlosgcampos: Add padding for future expansion here, since this struct is public

include/wpe-fdo/extensions/video-plane-display-dmabuf.h
```diff
@@ -0,0 +1,56 @@
...
+ * HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
```
carlosgcampos: Same here

src/extensions/video-plane-display-dmabuf-receiver.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+#include "ws.h"
+#include <unistd.h>
+
+static struct {
+    const struct wpe_video_plane_display_dmabuf_receiver* receiver;
+    void* data;
+} s_registered_receiver;
+
+static void dispatchReceiver(struct wpe_video_plane_display_dmabuf_export* dmabuf_export, uint32_t id, int fd, int32_t x, int32_t y, int32_t width, int32_t height, uint32_t stride)
```
carlosgcampos: This is a cpp file, this could be a lambda.

include/wpe-fdo/extensions/video-plane-display-dmabuf-receiver.h
```diff
@@ -0,0 +1,53 @@
...
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct wpe_video_plane_display_dmabuf_export;
+
+struct wpe_video_plane_display_dmabuf_receiver {
+    void (*handle_dmabuf)(void* data, struct wpe_video_plane_display_dmabuf_export*, uint32_t id, int fd, int32_t x, int32_t y, int32_t width, int32_t height, uint32_t stride);
```
carlosgcampos: why are x, y, w, h int32_t and not uint32_t?

src/extensions/video-plane-display-dmabuf.cpp
```diff
@@ -0,0 +1,346 @@
...
+    static DmaBufThread& singleton();
+    static void initialize(struct wl_display*);
+
+    struct wl_event_queue* eventQueue() const { return m_wl.eventQueue; }
+
+private:
+    static DmaBufThread* s_thread;
+    static gpointer s_threadEntrypoint(gpointer);
+
+    DmaBufThread(struct wl_display*);
```
carlosgcampos: explicit

src/extensions/video-plane-display-dmabuf.cpp
```diff
@@ -0,0 +1,346 @@
...
+        auto& source = *reinterpret_cast<PollingSource*>(base);
+
+        if (source.isReading) {
+            wl_display_cancel_read(source.display);
+            source.isReading = false;
+        }
+    },
+    nullptr, // closure_callback
+    nullptr, // closure_marshall
+};
```
carlosgcampos: This si duplicated from TargetSource::s_sourceFuncs, could we share that somehow?

src/ws.h
```diff
@@ -72,6 +74,11 @@ class Instance {
     const struct linux_dmabuf_buffer* getDmaBufBuffer(struct wl_resource*) const;
     void foreachDmaBufModifier(std::function<void (int format, uint64_t modifier)>);
 
+    using VideoPlaneDisplayDmaBufCallback = void(struct wpe_video_plane_display_dmabuf_export*, uint32_t, int, int32_t, int32_t, int32_t, int32_t, uint32_t);
+    void initializeVideoPlaneDisplayDmaBuf(VideoPlaneDisplayDmaBufCallback*);
```
carlosgcampos: Why not using std::function instead?

CMakeLists.txt
philn: The first header is used for the sender (WebKit) and the second for the consumer (Cog). Technically both headers could be merged I suppose though...

include/wpe-fdo/extensions/video-plane-display-dmabuf-receiver.h
philn: x and y at least can be negative (for surfaces partially offscreen). I just used these types to match the `wl_surface` API. Is there an issue?

carlosgcampos: Nah, it's ok

src/extensions/video-plane-display-dmabuf.cpp
philn: Seems so :)

philn:
This is ready for a second round. I think Zan disagrees about adding a new libwpe interface for this though.

zdobersek:
> This is ready for a second round. I think Zan disagrees about adding a new libwpe interface for this though.

The implementation exclusively depends on the dmabuf capability. Instead of baking this requirement into a libwpe interface, it makes more sense to me as an extension of sorts of a specific backend implementation that can enforce the dmabuf capability, which is the case for WPEBackend-fdo.

CMakeLists.txt
```diff
@@ -111,6 +120,18 @@ add_custom_command(
     VERBATIM
 )
 
+file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/video-plane-display-dmabuf)
+add_custom_command(
```
aperezdc: Instead of adding yet another `custom_command()` usage to invoke `wayland-scanner` it would be nicer to factor this out into a CMake function. For example Cog includes [in FindWaylandProtocols.cmake](https://github.com/Igalia/cog/blob/master/cmake/FindWaylandProtocols.cmake#L77) a CMake function which creates the needed targets and adds the generated files as part of the sources for a given target (note that it cannot be imported and used directly here, I mention it as something from where to draw inspiration).

In this case I think it's not a big deal to duplicate the `custom_command()`, but at some point it would be great to factor this out—specially if more protocols get added.

src/extensions/video-plane-display-dmabuf-receiver.cpp
```diff
@@ -0,0 +1,66 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "wpe-fdo/extensions/video-plane-display-dmabuf-receiver.h"
+
+#include "ws.h"
+#include <unistd.h>
+
+static struct {
```
aperezdc: This being static means that there can be only a single receiver. Somebody trying to use the API by reading the headers will not know about this and could potentially trip on this. It's not clear to me whether this needs to be (or not) a singleton, so I would like one of the two following changes done:

- Document the API (reminder: we have API reference generation in place using HotDoc!) and make it clear that there can be only one receiver, which is global, and therefore not multithread-safe.
- Change the API to remove the limitation of having only one global receiver.

In general we are lacking API documentation, so any effort towards adding and/or completing it would be greatly appreciated :wink:

src/ws.cpp
```diff
@@ -546,6 +599,50 @@ void Instance::foreachDmaBufModifier(std::function<void (int format, uint64_t mo
...
+
+void Instance::handleVideoPlaneDisplayDmaBuf(struct wpe_video_plane_display_dmabuf_export* dmabuf_export, uint32_t id, int fd, int32_t x, int32_t y, int32_t width, int32_t height, uint32_t stride)
+{
+    if (!m_videoPlaneDisplayDmaBuf.updateCallback) {
+        if (fd >= 0)
+            close(fd);
+        return;
+    }
+
+    (m_videoPlaneDisplayDmaBuf.updateCallback)(dmabuf_export, id, fd, x, y, width, height, stride);
```
aperezdc: The parentheses around `m_videoPlaneDisplayDmaBuf.updateCallback` are not needed, the following is fine:

```cpp
m_videoPlaneDisplayDmaBuf.updateCallback(dmabuf_export, id, fd, x, y, width, height, stride);
```

Either do that, or explicitly dereference the pointer when using parentheses:

```cpp
(*m_videoPlaneDisplayDmaBuf.updateCallback)(dmabuf_export, id, fd, x, y, width, height, stride);
```

(C++ and modern C don't require to explicitly dereference function pointers to invoke the function being pointed to, so I would rather go with removing the parentheses.)

src/ws.cpp
```diff
@@ -546,6 +599,50 @@ void Instance::foreachDmaBufModifier(std::function<void (int format, uint64_t mo
...
+}
+
+void Instance::handleVideoPlaneDisplayDmaBufEndOfStream(uint32_t id)
+{
+    if (!m_videoPlaneDisplayDmaBuf.endOfStreamCallback)
+        return;
+
+    (m_videoPlaneDisplayDmaBuf.endOfStreamCallback)(id);
```
aperezdc: Same here.

src/ws-client.cpp
```diff
@@ -167,21 +184,8 @@ void BaseTarget::initialize(struct wl_display* display)
...
-        source.display = display;
-        source.queue = m_wl.eventQueue;
-        source.isReading = false;
-
-        g_source_add_poll(m_glib.wlSource, &source.pfd);
-        g_source_set_name(m_glib.wlSource, "WPEBackend-fdo::wayland");
-        g_source_set_can_recurse(m_glib.wlSource, TRUE);
-        g_source_attach(m_glib.wlSource, g_main_context_get_thread_default());
-    }
+    m_glib.wlSource = ws_polling_source_new("WPEBackend-fdo::wayland", display, m_wl.eventQueue);
```
aperezdc: Nice refactor :smile:

CMakeLists.txt
aperezdc: In other parts of the lower-level WPE stack (like in `libwpe`) we use a single header in these cases, so I would vote for keeping this consistent with the existing practice.

src/extensions/video-plane-display-dmabuf-receiver.cpp
philn: I have added some docs :)

CMakeLists.txt
philn: Is it acceptable to do this as a follow-up PR?

philn:
The `DmaBufThread` indeed seems un-necessary. Thanks for pointing it out, Adrían!

philn:
Ok, I spoke too soon. The thread is actually not needed per-se, but without it, the browser process might get stuck while handling JS, for instance, as Zan pointed out.

aperezdc:
> Ok, I spoke too soon. The thread is actually not needed per-se, but without it, the browser process might get stuck while handling JS, for instance, as Zan pointed out.

Right, so to guarantee that updates are passed around steadily without jitter from other things running in the main thread's event loop. Let's merge this.

PR #89, opened 2020-01-04
Make WS::Instance::initialize crash on failure

mcatanzaro:
We need crashes here to debug
https://bugs.webkit.org/show_bug.cgi?id=201507. Otherwise, we're never
going to figure out what is going wrong. I decided to use g_error()
rather than g_warning() because we are going to want to be able to
examine stack and member variables to see as much of what went wrong as
possible. Having a backtrace will be much more useful for debugging than
just having a warning.

But because the function is designed to return false on failure, and
this commit subverts that, we will eventually want to revert this,
either turning the crashes into warnings or just removing the warnings
entirely. I've added a FIXME so we don't forget to do this.

### DIFF
--- a/src/ws.cpp
+++ b/src/ws.cpp
@@ -336,11 +336,17 @@ static bool isEGLExtensionSupported(const char* extensionList, const char* exten
 
 bool Instance::initialize(EGLDisplay eglDisplay)
 {
+    /* FIXME: Replace g_error() with g_warning() after https://bugs.webkit.org/show_bug.cgi?id=201507
+     * has been debugged and resolved. g_error() does not return, so the "return false" statements
+     * below are all currently unreachable dead code. They are kept so that you can simply replace
+     * each g_error() with g_warning() in the future without otherwise modifying the code.
+     */
+
     if (m_eglDisplay == eglDisplay)
         return true;
 
     if (m_eglDisplay != EGL_NO_DISPLAY) {
-        g_warning("Multiple EGL displays are not supported.\n");
+        g_error("Failed to initialize WS::Instance: multiple EGL displays are not supported.");
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


mcatanzaro:
> Or is there some reason to not do that?

Yeah: that will reduce the number of users reporting crashes down to just ourselves, because nobody else is going to set that. If we want to fix the bug we need real crash reports.

aperezdc:
> > Or is there some reason to not do that?
> 
> Yeah: that will reduce the number of users reporting crashes down to just ourselves, because nobody else is going to set that. If we want to fix the bug we need real crash reports.

Right, you make a good point. Then indeed it may be good to have this use `g_error()` at least during the development cycle, and dialing it back to `g_warning()` for the upcoming 1.6.x releases.

mcatanzaro:
@aperezdc, help. :) After we fix https://bugs.webkit.org/show_bug.cgi?id=209118, then https://bugs.webkit.org/show_bug.cgi?id=201507 will be the next major crash that we need to debug. We have no chance if this remains stalled.

mcatanzaro:
BTW this might be dangerous, since we only want to introduce crashes if there is an actual programmer error condition, not in expected failure situations. I know this introduces crashes in flatpak apps that don't have DRI access, for example. Ideally someone who understands EGL would take a hard look at this MR and at https://bugs.webkit.org/show_bug.cgi?id=201507 and decide where we should be adding debugging to catch the problem.

mcatanzaro:
> BTW this might be dangerous, since we only want to introduce crashes if there is an actual programmer error condition, not in expected failure situations.

Erm, debugging https://bugs.webkit.org/show_bug.cgi?id=209118, it looks like this change would make WebKit always crash when in VMs. So that's really bad. Don't merge.

I'm not closing this, though, because we do need to figure out what debug we can get away with adding.

mcatanzaro:
Let's continue in https://bugs.webkit.org/show_bug.cgi?id=201507

PR #97, opened 2020-02-04
Remove eglmesaext.h

JoshuaWatt:
eglmesaext.h is not provided by all EGL implementations, but those
implementations typically provide the necessary definitions in eglext.h.
Add a configure time check to see if eglmesaext.h is present and include
it if so.

### DIFF
--- a/src/ws.cpp
+++ b/src/ws.cpp
@@ -27,7 +27,6 @@
 
 #include <EGL/egl.h>
 #include <EGL/eglext.h>
-#include <EGL/eglmesaext.h>
 #include "linux-dmabuf/linux-dmabuf.h"
 #include "bridge/wpe-bridge-server-protocol.h"
 #include "video-plane-display-dmabuf/wpe-video-plane-display-dmabuf-server-protocol.h"
@@ -40,6 +39,7 @@
 #ifndef EGL_WL_bind_wayland_display
 #define EGL_WAYLAND_BUFFER_WL 0x31D5
 typedef EGLBoolean (EGLAPIENTRYP PFNEGLBINDWAYLANDDISPLAYWL) (EGLDisplay dpy, struct wl_display *display);
+typedef EGLBoolean (EGLAPIENTRYP PFNEGLQUERYWAYLANDBUFFERWL) (EGLDisplay dpy, struct wl_resource *buffer, EGLint attribute, EGLint *value);
 #endif
 
 #ifndef EGL_EXT_image_dma_buf_import_modifiers


philn:
LGTM. Any thoughts @aperezdc ?

src/ws.cpp
```diff
@@ -27,7 +27,9 @@
 
 #include <EGL/egl.h>
 #include <EGL/eglext.h>
-#include <EGL/eglmesaext.h>
+#ifdef HAVE_EGLMESAEXT_H
+#   include <EGL/eglmesaext.h>
```
philn: Nit: we usually don't use spaces after `#`

aperezdc:
I think we need to do the check a bit smarter; consider a build scenario where:

- Headers for two EGL implementations are installed; Mesa being one of them.
- The Mesa EGL headers are located somewhere in the default search path like `/usr/include`.
- Configuration passes `EGL_INCLUDE_DIR` and `EGL_LIBRARY` to CMake pointing to the non-Mesa implementation; for example: `cmake -DEGL_INCLUDE_DIR=/opt/somevendor/include -DEGL_LIBRARY=/opt/somevendor/lib`.

In this case we do not want to include `eglmesaext.h` from `/usr/include/EGL/eglmesaext.h` even if it is available, because the CMake invocation is trying to explicitly use an EGL implementation located elsewhere—mixing headers provided by two different EGL implementations sounds like a recipe for painful debugging.

What I would do is modifying `cmake/FindEGL.cmake`, using `find_path` to explicitly check whether the header is available in the configured (or detected) headers path:

```cmake
find_path(EGL_HAS_EGLMESAEXT
    NAMES EGL/eglmesaext.h
    PATHS ${EGL_INCLUDE_DIR}
    NO_DEFAULT_PATH
)
if (EGL_HAS_EGLMESAEXT)
    set(EGL_HAS_EGLMESAEXT ON)
else ()
    set(EGL_HAS_EGLMESAEXT OFF)
endif ()
```

@JoshuaWatt: Would you like to try adding the above snippet to `FindEGL.cmake` to determine whether the EGL implementation being used has the header? Let me know if you have any doubt, I'll be happy to help.

zdobersek:
IMO we should not include `eglmesaext.h` and instead just provide our own definitions.

JoshuaWatt:
Ping?

aperezdc:
This was also merged to the `wpebackend-fdo-1.4` branch as commit 58c54c1f17d0d1373485329ca9824deb0d115ddc — thanks @JoshuaWatt!

PR #99, opened 2020-02-25
Install to the libwpe backends directory

aperezdc:
Use pkg-config to obtain the value of the `backendsdir` variable to use the value as location for installing the backend. This also changes the pkg-config `.pc` file to have that list that directory in the `Libs` variable, so  pplications which need to link into `libWPEBackend-fdo-1.0` will still be able to find it.

----

This needs the changes from https://github.com/WebPlatformForEmbedded/libwpe/pull/64 and therefore the CI is expected to fail. See  https://github.com/WebPlatformForEmbedded/libwpe/issues/59 for more details.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,7 +153,7 @@ set_target_properties(WPEBackend-fdo
 
 install(
     TARGETS WPEBackend-fdo
-    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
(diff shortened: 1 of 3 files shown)


mcatanzaro (approved):

septatrix:
Any progress on this in the last 3.5 years? It is quite annoying to have to create that symlink locally just for the browser to work

aperezdc:
@septatrix Sorry, no progress on this, it's been very low priority and we are worried that it might affect existing deployments. Therefore, we will not be merging this for the time being.

PR #101, opened 2020-03-02
ws: Do not block on wl_event_loop_dispatch()

### LINKED ISSUE #94: Hanging in wl_event_loop_dispatch in WebKitGTK+WPE on FreeBSD
Since WebKitGTK started using WPE, WebKitGTK apps were not working on Wayland on FreeBSD: the application would be frozen with no content rendered.
I've tried to debug today, and turns out just attaching the debugger and continuing unfreezes the application and everything works.

Where apps get frozen is in the `wl_event_loop_dispatch` that's called with no timeout (`-1`):

```
  * frame #0: 0x0000000806e957ca libc.so.7`_kevent + 10
    frame #1: 0x000000080316a783 libthr.so.3`___lldb_unnamed_symbol54$$libthr.so.3 + 83
    frame #2: 0x000000080862aafb libepoll-shim.so.0`epoll_wait + 315
    frame #3: 0x0000000806c3e9fd libwayland-server.so.0`wl_event_loop_dispatch + 77
    frame #4: 0x0000000807145128 libWPEBackend-fdo-1.0.so.1`___lldb_unnamed_symbol35$$libWPEBackend-fdo-1.0.so.1 + 4
    frame #5: 0x0000000803cec267 libglib-2.0.so.0`g_main_context_dispatch + 327
```

I see that this is only called when polling has returned that there is something to read, but somehow it's being called when it has nothing to read. Maybe GTK's Wayland dispatch call happened in between the source receiving the poll result and dispatch being called.

aperezdc:
Instruct `wl_display_get_event_loop()` to not block by passing zero as timeout, instead of letting it block indefinitely.

Fixes #94

### DIFF
--- a/src/ws.cpp
+++ b/src/ws.cpp
@@ -117,7 +117,7 @@ GSourceFuncs ServerSource::s_sourceFuncs = {
 
         if (source.pfd.revents & G_IO_IN) {
             struct wl_event_loop* eventLoop = wl_display_get_event_loop(source.display);
-            wl_event_loop_dispatch(eventLoop, -1);
+            wl_event_loop_dispatch(eventLoop, 0);
             wl_display_flush_clients(source.display);
         }
 


aperezdc:
Cherry-picked to the `wpebackend-fdo-1.6` branch as commit 970ef95d2c3bb5161178a0be267b3b10d19b1bee :+1:

PR #105, opened 2020-03-03
CMake: Use "hidden" symbol visibility by default

### LINKED ISSUE #103: Lots of internal symbols exported
WPEBackend-fdo is exporting lots of symbols that are supposed to be internal. Anything beginning with Impl:: or WS::, for instance.

Adrian says: "We should switch to `-fvisibility=hidden` and explicitly mark public symbols." I agree.

Would make sense to make sure libwpe is in good shape too.

aperezdc:
Arrange passing `-fvisibility=hidden` to the compiler where supported, which effectively disables exporting symbols which are not explicitly marked as visible.

No further changes are needed because the public API functions were already marked with `__attribute__((visibility("default")))`, which marks them as exported.

Fixes #103


mcatanzaro:
Should do this for libwpe too, yes?

aperezdc:
> Should do this for libwpe too, yes?

Wouldn't hurt. Somehow I thought it was already being done, but I was wrong... Let's cook up a patch!

aperezdc:
Cherry-picked into the `wpebackend-fdo-1.6` branch as commit 1bc41a5cb4ec288c2e166249e9d5b2db0f73a2a6 :construction_worker_man:

aperezdc:
> > Should do this for libwpe too, yes?
> 
> Wouldn't hurt. Somehow I thought it was already being done, but I was wrong... Let's cook up a patch!

@mcatanzaro See https://github.com/WebPlatformForEmbedded/libwpe/pull/65 for the libwpe patch :wink:

PR #107, opened 2020-03-10
Provide impl facilities for WS::Instance

zdobersek:
Have the WS::Instance class adopt an implementation object that extends
the WS::Instance::Impl interface. This would allow for different types
of implementations being instantiated on the Wayland IPC channel that's
managed by WS::Instance.

The current, EGL-oriented implementation is moved into the ImplEGL
class, and the existing interfaces are made to implicitly depend on
that implementation being used -- which still works since it's the
only implementation currently available. The integration of specific
export interfaces along with the viable implementations will be
improved once more implementations are added.


carlosgcampos:
LGTM

PR #108, opened 2020-03-11
doc: Fix HotDoc warnings about missing headers

aperezdc:
Add the missing include directories to the HotDoc command and arrange headers to use local inclusion (with `""` instead of `<>`) to ensure that source tree files are picked first.


aperezdc:
Cherry-picked into the `wpebackend-fdo-1.6` branch as commit `f1a31aeadce2f79052cd3e9686a4a688e121afd1`.

PR #112, opened 2020-04-06
Use libepoxy for EGL operations

zdobersek:
Instead of relying on EGL headers and library provided at compile-time,
use libepoxy to push that necessity to runtime.

The CMake integration follows the new 'imported' model. Instead of
manually parsing the EGL extension list to detect present extensions,
we can rely on the epoxy_has_egl_extension() function.

### DIFF
--- a/src/ws-egl.cpp
+++ b/src/ws-egl.cpp
@@ -26,8 +26,7 @@
 #include "ws-egl.h"
 
 #include "linux-dmabuf/linux-dmabuf.h"
-#include <EGL/egl.h>
-#include <EGL/eglext.h>
+#include <epoxy/egl.h>
 #include <cassert>
 #include <cstring>
 
@@ -77,21 +76,6 @@ static PFNEGLQUERYDMABUFFORMATSEXTPROC s_eglQueryDmaBufFormatsEXT;
 static PFNEGLQUERYDMABUFMODIFIERSEXTPROC s_eglQueryDmaBufModifiersEXT;
 
 namespace WS {
-
-static bool isEGLExtensionSupported(const char* extensionList, const char* extension)
-{
-    if (!extensionList)
-        return false;
-
-    int extensionLen = strlen(extension);
-    const char* extensionListPtr = extensionList;
-    while ((extensionListPtr = strstr(extensionListPtr, extension))) {
-        if (extensionListPtr[extensionLen] == ' ' || extensionListPtr[extensionLen] == '\0')
-            return true;
-        extensionListPtr += extensionLen;
-    }
-    return false;
-}
 
 ImplEGL::ImplEGL()
     : m_eglDisplay(EGL_NO_DISPLAY)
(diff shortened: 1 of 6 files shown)


src/ws-egl.cpp
```diff
@@ -180,8 +163,8 @@ bool ImplEGL::initialize(EGLDisplay eglDisplay)
     m_eglDisplay = eglDisplay;
 
     /* Initialize Linux dmabuf subsystem. */
-    if (isEGLExtensionSupported(extensions, "EGL_EXT_image_dma_buf_import")
-        && isEGLExtensionSupported(extensions, "EGL_EXT_image_dma_buf_import_modifiers")) {
+    if (epoxy_has_egl_extension(eglDisplay, "EGL_EXT_image_dma_buf_import")
+        && epoxy_has_egl_extension(eglDisplay, "EGL_EXT_image_dma_buf_import_modifiers")) {
         s_eglQueryDmaBufFormatsEXT = reinterpret_cast<PFNEGLQUERYDMABUFFORMATSEXTPROC>(eglGetProcAddress("eglQueryDmaBufFormatsEXT"));
```
aperezdc, line 168: When using libepoxy we do not need to use `eglGetProcAddress()` by hand for these functions either:

```
% grep 'define eglQueryDmaBuf' /usr/include/epoxy/* 
/usr/include/epoxy/egl_generated.h:#define eglQueryDmaBufFormatsEXT epoxy_eglQueryDmaBufFormatsEXT
/usr/include/epoxy/egl_generated.h:#define eglQueryDmaBufModifiersEXT epoxy_eglQueryDmaBufModifiersEXT
```

src/ws-egl.cpp
```diff
@@ -154,15 +138,14 @@ bool ImplEGL::initialize(EGLDisplay eglDisplay)
...
+    if (epoxy_has_egl_extension(eglDisplay, "EGL_WL_bind_wayland_display")) {
         s_eglBindWaylandDisplayWL = reinterpret_cast<PFNEGLBINDWAYLANDDISPLAYWL>(eglGetProcAddress("eglBindWaylandDisplayWL"));
         assert(s_eglBindWaylandDisplayWL);
         s_eglQueryWaylandBufferWL = reinterpret_cast<PFNEGLQUERYWAYLANDBUFFERWL>(eglGetProcAddress("eglQueryWaylandBufferWL"));
         assert(s_eglQueryWaylandBufferWL);
     }
 
-    if (isEGLExtensionSupported(extensions, "EGL_KHR_image_base")) {
+    if (epoxy_has_egl_extension(eglDisplay, "EGL_KHR_image_base")) {
         s_eglCreateImageKHR = reinterpret_cast<PFNEGLCREATEIMAGEKHRPROC>(eglGetProcAddress("eglCreateImageKHR"));
```
aperezdc, line 149: There two are defined by libepoxy, and we do not need to use `eglGetProcAddress()` for them:

```
% grep -E 'define egl(Create|Destroy)Image' /usr/include/epoxy/*
/usr/include/epoxy/egl_generated.h:#define eglCreateImage epoxy_eglCreateImage
/usr/include/epoxy/egl_generated.h:#define eglCreateImageKHR epoxy_eglCreateImageKHR
/usr/include/epoxy/egl_generated.h:#define eglDestroyImage epoxy_eglDestroyImage
/usr/include/epoxy/egl_generated.h:#define eglDestroyImageKHR epoxy_eglDestroyImageKHR
```

src/ws-egl.cpp
zdobersek: :+1:

src/ws-egl.cpp
zdobersek: :+1:

PR #118, opened 2020-05-15
meson: Declare dependency to be usable as subproject

thiblahute:


philn:
Even so, this currently can't be used as a subproject, because the header paths differ in the tree, from what's installed by `ninja install`. So I get this kind of error:
```
subprojects/gst-plugins-bad/ext/wpe/WPEThreadedView.h:25:10: fatal error: wpe/fdo.h: No such file or directory
```

philn:
LGTM :+1:

meson.build
```diff
@@ -211,21 +211,30 @@ elif not cxx.has_header('xkbcommon/xkbcommon-keysyms.h')
...
+lib = shared_library(meson.project_name() + '-' + api_version,
+	sources, generated_sources,
 	install: true,
 	dependencies: deps,
 	version: soversion,
 	soversion: soversion_major,
 	include_directories: include_directories('include'),
 )
 
+wpe_fdo_dep = declare_dependency(link_with : lib,
```
philn, line 232: perhaps `wpe_backend_fdo_dep` would be a better name?

aperezdc: The [documentation](https://mesonbuild.com/Subprojects.html) tells that the convention is `<project_name>_dep`. In our case the project name is `WPEBackend-fdo` which is not a valid variable name, so I think in this case it would be `wpebackend_fdo_dep`.

src/exported-image-egl.cpp
```diff
@@ -26,7 +26,7 @@
 #include "view-backend-exportable-fdo-egl-private.h"
 #include "ws.h"
 #include <cassert>
-#include <wpe-fdo/exported-image-egl.h>
+#include <wpe/exported-image-egl.h>
```
aperezdc, line 29: While at it, could we also change this to use `#include "..."` for consistency with the rest of inclusions of headers which are part of the source tree?

src/view-backend-exportable-fdo-egl.cpp
```diff
@@ -31,7 +31,7 @@
 #include <epoxy/egl.h>
 #include <cassert>
 #include <list>
-#include <wpe-fdo/view-backend-exportable-egl.h>
+#include <wpe/view-backend-exportable-egl.h>
```
aperezdc, line 34: Same here: `#include "..."`.

src/view-backend-exportable-private.h
```diff
@@ -28,7 +28,7 @@
 #include "ipc.h"
 #include "ws.h"
 #include <gio/gio.h>
-#include <wpe-fdo/view-backend-exportable.h>
+#include <wpe/view-backend-exportable.h>
```
aperezdc, line 31: And here: `#include "..."`

philn:
:100:

PR #119, opened 2020-05-16
renderer-backend-egl: cast return value in get_native_display()

zdobersek:
In the wpe_renderer_backend_egl_interface::get_native_display(),
explicitly cast the wl_display object to the EGLNativeDisplayType.
This should avoid compilation errors on rare EGL stacks that have
odd default EGLNativeDisplayType type definitions.


aperezdc:
Cherry-picked into the `wpebackend-fdo-1.6` branch as commit 925147d162a9326b428ad288da84955c6b0306c8.

PR #122, opened 2020-09-27
Audio extension fixes and additions

philn:


CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.3)
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 include(VersioningUtils)
 
-SET_PROJECT_VERSION(1 7 1)
+SET_PROJECT_VERSION(1 9 1)
```
aperezdc, line 6: Oh, I forgot to do this in `master` after branching for the 1.8.x stable series. Note that this needs to be `1 9 0` (it will be `1 9 1` in the commit for that particular release), and you need to modify the `CALCULATE_LIBRARY_VERSIONS_FROM_LIBTOOL_TRIPLE()` call to use `C+1 0 A+1`.

meson.build
```diff
@@ -7,7 +7,7 @@ project('WPEBackend-fdo', ['c', 'cpp'],
 		'cpp_std=c++11',
 	],
 	license: 'BSD-2-Clause',
-	version: '1.7.1',
+	version: '1.9.1',
```
aperezdc, line 10: Same here, it should be `1.9.0` and the libtool versioning numbers below need to be the same as in `CMakeLists.txt`.

aperezdc (changes requested):
Code looks fine. It would be great if you could apply a small change to the patch that does the version bump before we merge this. 

Thanks for PR! :100:

CMakeLists.txt
philn: Thanks for review! Indeed I wasn't sure what to do with those numbers and if they should be updated at the beginning of the cycle or just before the final release, so I didn't touch them :)

aperezdc (approved):
This is perfect now, let's merge it :smile:

PR #127, opened 2020-11-07
view-backend-exportable-private: Add wl_client_add_destroy_listener in the ViewBackend

psaavedra:
**DESCRIPTION:**

This MR fixes a crash in the WebKit UI-process. This crash happens when the wl_client of the related to a surface of a blocked content loaded request is being closed because an WL_EVENT_HANGUP (5).

In this scenario, the UI-process doesn't know about this client being closed and because it has its own copy of the wl_client pointer it keeps trying to operate over it during the release/dispatch actions. Therefore, the client is being freed in the wayland code when the error happens but the ViewBackend will never will know that its copy of the pointer is pointing to an area that it will be deleted.

The WL_EVENT_HANGUP happens immediately after that the content is blocked. If we are able to close the client and/or unregister the surface related to this request in a civilized way, we could avoid this situation.

**HOW TO REPRODUCE the bug:**

The synthetic code to reproduce the crash is attached in the issue: [launcher.c](https://bugs.webkit.org/attachment.cgi?id=413286)

* Build instructions:

```
# Set this variables to the right place where you have the WPE related libraries
# export PKG_CONFIG_PATH=/home/igalia/psaavedra/install/lib/pkgconfig/
# export LD_LIBRARY_PATH=/home/igalia/psaavedra/install/lib

filename="launcher"

CFLAGS="gtk+-3.0 wpe-webkit-1.0 wpebackend-fdo-1.0 egl"
LIBS="gtk+-3.0 wayland-client wpe-1.0 wpebackend-fdo-1.0 wpe-web-extension-1.0  wpe-webkit-1.0 egl"

gcc -g3 `pkg-config --cflags $CFLAGS` -o $filename $filename.c `pkg-config --libs $LIBS`
```

* Execution: 

```
export WAYLAND_DISPLAY=wayland-0
export XDG_RUNTIME_DIR=/xdg
export LD_LIBRARY_PATH=/home/igalia/psaavedra/install/lib
export WEBKIT_IGNORE_SSL_ERRORS=1
./launcher
```

The issue is reproducible in:

* WebKit: 2.28; 2.30; trunk (2020-11-05)
* libwpe: 1.6; 1.8
* Weston: 1.3; 5.0 
* Wayland-protocols: 1.13 and 1.17

```
ii  wayland-protocols                      1.17-1                              all          wayland compositor protocols
ii  weston                                 5.0.0-3                             amd64        reference implementation of a wayland compositor
```

**SOLUTION:**

This MR is attempts to address this issue by:

* Add callback for wl_client_add_destroy_listener in the ViewBackend.This callback set the m_client to null.
* Guard the wl_client_flush() calls when wayland client object reference is null.

Ref: [[WPE] Crash if a view loads a blocked URL by a content filter - webkit#218612](https://bugs.webkit.org/show_bug.cgi?id=218612)

### DIFF
--- a/src/ws.cpp
+++ b/src/ws.cpp
@@ -532,6 +532,7 @@ void Instance::unregisterViewBackend(uint32_t surfaceId)
     auto it = m_viewBackendMap.find(surfaceId);
     if (it != m_viewBackendMap.end()) {
         it->second->apiClient = nullptr;
+        wl_client_destroy(it->second->client);
         m_viewBackendMap.erase(it);
     }
 }
--- a/src/view-backend-private.cpp
+++ b/src/view-backend-private.cpp
@@ -127,19 +127,19 @@ void ViewBackend::registerSurface(uint32_t surfaceId)
 {
     m_surfaceId = surfaceId;
     m_client = WS::Instance::singleton().registerViewBackend(m_surfaceId, *this);
-    this->m_destroyClientListener.notify = (wl_notify_func_t) [](struct wl_listener* listener, void* data)
+ 
+    struct wl_client_destroy_listener *listener = new wl_client_destroy_listener {this, };
+    listener->destroyClientListener.notify = (wl_notify_func_t) [](struct wl_listener* listener, void* data)
     {
-        ViewBackend *viewBackend;
-        viewBackend = wl_container_of(listener, viewBackend, m_destroyClientListener);
+        struct wl_client_destroy_listener *container;
+        container = wl_container_of(listener, container, destroyClientListener);
 
         struct wl_client* client = (struct wl_client*) data;
-        g_debug("ViewBackend <%p>: wl_client <%p> destroy notification for fd %d", viewBackend, data, wl_client_get_fd(client));
-        viewBackend->m_client = NULL;
+        container->backend->m_client = NULL;
+        delete container;  // Release the wl_client_destroy_listener instance since this is not longer needed.
     };
     wl_client_add_destroy_listener(m_client,
-                                   &this->m_destroyClientListener);
-
-
+                                   &listener->destroyClientListener);
 }
 
 void ViewBackend::unregisterSurface(uint32_t surfaceId)
--- a/src/view-backend-private.cpp
+++ b/src/view-backend-private.cpp
@@ -111,35 +111,25 @@ void ViewBackend::dispatchFrameCallbacks()
     }
     clearFrameCallbacks();
 
-    if (m_client)
-        wl_client_flush(m_client);
+    if (m_client.object)
+        wl_client_flush(m_client.object);
     wpe_view_backend_dispatch_frame_displayed(m_backend);
 }
 
 void ViewBackend::releaseBuffer(struct wl_resource* buffer_resource)
 {
     wl_buffer_send_release(buffer_resource);
-    if (m_client)
-        wl_client_flush(m_client);
+    if (m_client.object)
+        wl_client_flush(m_client.object);
 }
 
 void ViewBackend::registerSurface(uint32_t surfaceId)
 {
     m_surfaceId = surfaceId;
-    m_client = WS::Instance::singleton().registerViewBackend(m_surfaceId, *this);
- 
-    struct wl_client_destroy_listener *listener = new wl_client_destroy_listener {this, };
-    listener->destroyClientListener.notify = (wl_notify_func_t) [](struct wl_listener* listener, void* data)
-    {
-        struct wl_client_destroy_listener *container;
-        container = wl_container_of(listener, container, destroyClientListener);
+    m_client.object = WS::Instance::singleton().registerViewBackend(m_surfaceId, *this);
 
-        struct wl_client* client = (struct wl_client*) data;
-        container->backend->m_client = NULL;
-        delete container;  // Release the wl_client_destroy_listener instance since this is not longer needed.
-    };
-    wl_client_add_destroy_listener(m_client,
-                                   &listener->destroyClientListener);
+    m_client.destroyListener.notify = Client::destroyNotify;
+    wl_client_add_destroy_listener(m_client.object, &m_client.destroyListener);
 }
 
 void ViewBackend::unregisterSurface(uint32_t surfaceId)
@@ -178,6 +168,14 @@ void ViewBackend::clearFrameCallbacks()
         delete resource;
     }
     wl_list_init(&m_frameCallbacks);
+}
+
+void ViewBackend::Client::destroyNotify(struct wl_listener* listener, void*)
+{
+    Client* client;
+    client = wl_container_of(listener, client, destroyListener);
+
+    client->object = nullptr;
 }
 
 void ViewBackend::FrameCallbackResource::destroyNotify(struct wl_listener* listener, void*)
--- a/src/view-backend-private.cpp
+++ b/src/view-backend-private.cpp
@@ -111,20 +111,35 @@ void ViewBackend::dispatchFrameCallbacks()
     }
     clearFrameCallbacks();
 
-    wl_client_flush(m_client);
+    if (m_client)
+        wl_client_flush(m_client);
     wpe_view_backend_dispatch_frame_displayed(m_backend);
 }
 
 void ViewBackend::releaseBuffer(struct wl_resource* buffer_resource)
 {
     wl_buffer_send_release(buffer_resource);
-    wl_client_flush(m_client);
+    if (m_client)
+        wl_client_flush(m_client);
 }
 
 void ViewBackend::registerSurface(uint32_t surfaceId)
 {
     m_surfaceId = surfaceId;
     m_client = WS::Instance::singleton().registerViewBackend(m_surfaceId, *this);
+    this->m_destroyClientListener.notify = (wl_notify_func_t) [](struct wl_listener* listener, void* data)
+    {
+        ViewBackend *viewBackend;
+        viewBackend = wl_container_of(listener, viewBackend, m_destroyClientListener);
+
+        struct wl_client* client = (struct wl_client*) data;
+        g_debug("ViewBackend <%p>: wl_client <%p> destroy notification for fd %d", viewBackend, data, wl_client_get_fd(client));
+        viewBackend->m_client = NULL;
+    };
+    wl_client_add_destroy_listener(m_client,
+                                   &this->m_destroyClientListener);
+
+
 }
 
 void ViewBackend::unregisterSurface(uint32_t surfaceId)
--- a/src/view-backend-private.h
+++ b/src/view-backend-private.h
@@ -100,7 +100,11 @@ private:
 
     std::unique_ptr<FdoIPC::Connection> m_socket;
     int m_clientFd { -1 };
-    struct wl_listener m_destroyClientListener;
+};
+
+struct wl_client_destroy_listener {
+    ViewBackend* backend;
+    struct wl_listener destroyClientListener;
 };
 
 struct wpe_view_backend_private {
--- a/src/view-backend-private.h
+++ b/src/view-backend-private.h
@@ -100,6 +100,7 @@ private:
 
     std::unique_ptr<FdoIPC::Connection> m_socket;
     int m_clientFd { -1 };
+    struct wl_listener m_destroyClientListener;
 };
 
 struct wpe_view_backend_private {
--- a/src/view-backend-private.h
+++ b/src/view-backend-private.h
@@ -91,7 +91,12 @@ private:
     static gboolean s_socketCallback(GSocket*, GIOCondition, gpointer);
 
     uint32_t m_surfaceId { 0 };
-    struct wl_client* m_client { nullptr };
+    struct Client {
+        struct wl_client* object { nullptr };
+        struct wl_listener destroyListener;
+
+        static void destroyNotify(struct wl_listener*, void*);
+    } m_client;
 
     ClientBundle* m_clientBundle;
     struct wpe_view_backend* m_backend;
@@ -100,11 +105,6 @@ private:
 
     std::unique_ptr<FdoIPC::Connection> m_socket;
     int m_clientFd { -1 };
-};
-
-struct wl_client_destroy_listener {
-    ViewBackend* backend;
-    struct wl_listener destroyClientListener;
 };
 
 struct wpe_view_backend_private {


src/view-backend-exportable-private.h
```diff
@@ -102,6 +102,11 @@ class ViewBackend final : public WS::ExportableClient, public FdoIPC::MessageRec
     int m_clientFd { -1 };
 };
 
+struct wl_client_destroy_listener {
+    ViewBackend* backend;
+    struct wl_listener destroyClientListener;
```
aperezdc, line 107: Is there any reason why the `wl_listener` is not a member of the `ViewBackend` type?

src/view-backend-exportable-private.cpp
```diff
@@ -111,20 +111,36 @@ void ViewBackend::dispatchFrameCallbacks()
...
     m_client = WS::Instance::singleton().registerViewBackend(m_surfaceId, *this);
+ 
+    struct wl_client_destroy_listener *listener = new wl_client_destroy_listener {this, };
+    listener->destroyClientListener.notify = (wl_notify_func_t) [](struct wl_listener* listener, void* data)
+    {
+        struct wl_client_destroy_listener *container;
+        container = wl_container_of(listener, container, destroyClientListener);
+
+        struct wl_client* client = (struct wl_client*) data;
+        g_debug("ViewBackend <%p>: wl_client <%p> destroy notification received for fd %d", container->backend, data, wl_client_get_fd(client));
```
aperezdc, line 138: I suppose we don't want to keep this debugging code in the final version of the patch, right?

aperezdc (changes requested):
This seems to have merge conflicts, so at least a rebase will be needed. I have written a couple of comments as well.

src/view-backend-exportable-private.cpp
psaavedra: Yes, makes senses.

src/view-backend-exportable-private.h
psaavedra: This was my first try but I got in crashes in situations where the ViewBackend is destroyed before the wl_client client is deleted. That is the normal case where the ViewBackend is destroyed because the WebProcess is not longer needed (elimination of all the views related to one WebContext). In that scenario when the wl_callback for the listener is called the memory where the wl_listener is is already tagged as freed so unexpected behavior can happens. In any case we were using a memory area already freed.

That is the reason why I moved the callback to an auxiliary structure which is attached to the wl_client_add_destroy_listener and it contains a reference to the ViewBackend. This instance is erased in the destroy notification callback. 

In other case why are going to experiment this crash:


```
#1  0x70683b6c in wl_priv_signal_emit (signal=signal@entry=0x6f3d2650, data=data@entry=0x6f3d2610) at /root/sources/wayland/src/wayland-server.c:2024
#2  0x70683d00 in wl_client_destroy (client=client@entry=0x6f3d2610) at /root/sources/wayland/src/wayland-server.c:866
#3  0x70683e8c in wl_client_connection_data (fd=64, mask=<optimized out>, data=0x6f3d2610) at /root/sources/wayland/src/wayland-server.c:342
#4  0x70685d10 in wl_event_loop_dispatch (loop=0x6f310e38, timeout=timeout@entry=0) at /root/sources/wayland/src/event-loop.c:426
#5  0x6f4c69b8 in WS::ServerSource::<lambda(GSource*, GSourceFunc, gpointer)>::operator() (__closure=0x0, base=0x6f3110f0) at /root/sources/sources/wpebackend-fdo/src/ws.cpp:120
#6  WS::ServerSource::<lambda(GSource*, GSourceFunc, gpointer)>::_FUN(GSource *, GSourceFunc, gpointer) () at /root/sources/sources/wpebackend-fdo/src/ws.cpp:129
#7  0x73f76668 in g_main_context_dispatch () from /root/usr/lib/libglib-2.0.so.0.5800.3
#8  0x73f7688c in ?? () from /root/usr/lib/libglib-2.0.so.0.5800.3
#9  0x73f76c2c in g_main_loop_run () from /root/usr/lib/libglib-2.0.so.0.5800.3
#10 0x004e06f8 in wpe::Launcher::runThread (this=this@entry=0x21b19c0) at /root/sources/sources/wpe-launcher/src/Launcher.cc:895
#11 0x004eb414 in wpe::Launcher::<lambda(void*)>::operator() (__closure=0x0, data=<optimized out>) at /root/sources/sources/wpe-launcher/src/Launcher.cc:499
#12 wpe::Launcher::<lambda(void*)>::_FUN(void *) () at /root/sources/sources/wpe-launcher/src/Launcher.cc:503
#13 0x73fa2bdc in ?? () from /root/usr/lib/libglib-2.0.so.0.5800.3
#14 0x766ebe8c in ?? () from /root/lib/libpthread-2.26.so
```


Where there `wl_priv_signal_emit()` fails for this reason:

```
void
wl_priv_signal_emit(struct wl_priv_signal *signal, void *data)
{
        struct wl_listener *l;
        struct wl_list *pos;

        wl_list_insert_list(&signal->emit_list, &signal->listener_list);
        wl_list_init(&signal->listener_list);

        /* Take every element out of the list and put them in a temporary list.
         * This way, the 'it' func can remove any element it wants from the list
         * without troubles, because we always get the first element, not the
         * one after the current, which may be invalid.
         * wl_list_for_each_safe tries to be safe but it fails: it works fine
         * if the current item is removed, but not if the next one is. */
        while (!wl_list_empty(&signal->emit_list)) {
                pos = signal->emit_list.next;
                l = wl_container_of(pos, l, link);
                fprintf(stderr, "XXX wl_priv_signal_emit: l->notify: %p \n", l->notify);  // <<<<<<< Shows a memory address different to the memory where it should be the struct wl_listener pointer associated to the ViewBackend when this was already deleted

                wl_list_remove(pos);
                wl_list_insert(&signal->listener_list, pos);

                l->notify(l, data);  // <<<<<<< Crash
        }
}
```

psaavedra:
> This seems to have merge conflicts, so at least a rebase will be needed. I have written a couple of comments as well.

Rebased.

src/view-backend-exportable-private.h
aperezdc: The destructor for `ViewBackend` needs to do this to unregister the listener for the destruction signal. It is not obvious because “apparently” there is no API for that, but reading the documentation for `wl_signal` one can find this hidden gem (emphasis mine):

> […] Observers are `wl_listener`'s that are added through `wl_signal_add`. Signals are emitted using `wl_signal_emit`, which will invoke all listeners <strong>until that listener is removed by `wl_list_remove()`</strong> (or whenever the signal is destroyed).

So having something like this:

```cpp
class ViewBackend final : // ... 
{
private:
    // ...
    struct wl_listener m_destroyListener { };
};
```

I would actually add the call to remove the listener in `ViewBackend::unregisterSurface()`, which is also called by the `~ViewBackend()` destructor, but can be called before the instance is deleted, in response to an IPC message as well.
 
```cpp
void ViewBackend::unregisterSurface(uint32_t surfaceId)
{
    if (!surfaceId || m_surfaceId != surfaceId)
        return;

    wl_list_remove(&m_destroyListener);  // Added!
    clearFrameCallbacks();

    WS::Instance::singleton().unregisterViewBackend(m_surfaceId);
    m_surfaceId = 0;
}
```

I would initialize `m_destroyListener.notify` inside the class constructor: it can be only once because the function assigned is always the same and it won't change, so there is no need to initialize it every time that `ViewBackend::registerSurface()` is called. The call to register the listener with `wl_client_add_destroy_listener()` would still be done in `ViewBackend::registerSurface()`, of course.

psaavedra: `wl_list_remove()`  ... interesting ... actually is what the `wl_priv_signal_emit()` function does into the while. The question  is, that function belongs to the `wayland/src/wayland-private.h` header. I :thinking: if it is good idea to use such type of calls outside the wayland code.

aperezdc: Nononono. I am telling you to use `wl_list_remove()`, which is declared is in the **public** `wayland-util.h` header.

psaavedra: Ooops. I bad grep-ing before. Yes you are right:


```
../wayland/src/wayland-server-core.h: * If you need to remove a listener from a signal, use wl_list_remove().
../wayland/src/wayland-server-core.h: * wl_list_remove(&your_listener.link);
../wayland/src/wayland-server-core.h: * listener is removed by wl_list_remove() (or whenever the signal is
../wayland/src/wayland-util.h: * safety by invoking wl_list_init() before wl_list_remove().
../wayland/src/wayland-util.h:wl_list_remove(struct wl_list *elm);
```

Looks like a true candidate.

src/view-backend-exportable-private.cpp
psaavedra: I've cherry-picked the cleanups suggested by @zdobersek. Also, I0ve verified the `unregisterViewBackend()` is called when the  WebView is destroyed (ex: `g_object_unref()`) so It looks like the right place to put the call for the  `wl_client_destroy()`.

psaavedra:
I've cherry-picked the cleanups suggested by @zdobersek. Also, I've verified the `unregisterViewBackend()` is called when the  WebView is destroyed (ex: `g_object_unref()`) so It looks like the right place to put the call for the  `wl_client_destroy()`.

psaavedra:
Should we squashed all together in just one single commit before merge?

aperezdc:
@psaavedra: Ping! Could you please rebase this MR, if it is still relevant? Thanks!

aperezdc (approved):
Now that I managed to fix the CI issue (it had nothing to do with your changes), this looks good to go :rocket: 

Thanks!

src/ws.cpp
```diff
@@ -532,6 +532,7 @@ void Instance::unregisterViewBackend(uint32_t surfaceId)
     auto it = m_viewBackendMap.find(surfaceId);
     if (it != m_viewBackendMap.end()) {
         it->second->apiClient = nullptr;
+        wl_client_destroy(it->second->client);
```
csaavedra, line 535: I'm seeing a crash in this call that leads me to believe that this client has already been destroyed elsewhere when this is called. Here is the stacktrace:

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  wl_list_insert_list (list=list@entry=0x6f00b528, other=other@entry=0x6f00b520) at /usr/src/debug/wayland/1.13.0-r0/wayland-1.13.0/src/wayland-util.c:89
89            other->next->prev = list;
[Current thread is 1 (LWP 16)]
(gdb) back
#0  wl_list_insert_list (list=list@entry=0x6f00b528, other=other@entry=0x6f00b520) at /usr/src/debug/wayland/1.13.0-r0/wayland-1.13.0/src/wayland-util.c:89
#1  0x703b8ac4 in wl_priv_signal_emit (signal=signal@entry=0x6f00b520, data=data@entry=0x6f00b4e0) at /usr/src/debug/wayland/1.13.0-r0/wayland-1.13.0/src/wayland-server.c:1992
#2  0x703b8c70 in wl_client_destroy (client=0x6f00b4e0) at /usr/src/debug/wayland/1.13.0-r0/wayland-1.13.0/src/wayland-server.c:852
#3  0x6fb140d4 in WS::Instance::unregisterViewBackend (this=0x6f01d530, surfaceId=1) at /usr/src/debug/wpebackend-fdo/1.6.1-r0/wpebackend-fdo-1.6.1/src/ws.cpp:673
#4  0x6fb123a4 in ViewBackend::unregisterSurface (this=this@entry=0x6f0b37c0, surfaceId=<optimized out>)
    at /usr/src/debug/wpebackend-fdo/1.6.1-r0/wpebackend-fdo-1.6.1/src/view-backend-exportable-private.cpp:137
#5  0x6fb124bc in ViewBackend::unregisterSurface (surfaceId=<optimized out>, this=0x6f0b37c0) at /usr/src/debug/wpebackend-fdo/1.6.1-r0/wpebackend-fdo-1.6.1/src/view-backend-exportable-private.cpp:49
#6  ViewBackend::~ViewBackend (this=0x6f0b37c0, __in_chrg=<optimized out>) at /usr/src/debug/wpebackend-fdo/1.6.1-r0/wpebackend-fdo-1.6.1/src/view-backend-exportable-private.cpp:45
#7  0x6fb15e74 in <lambda(void*)>::operator() (__closure=<optimized out>, data=<optimized out>) at /usr/src/debug/wpebackend-fdo/1.6.1-r0/wpebackend-fdo-1.6.1/src/view-backend-exportable-private.h:123
#8  <lambda(void*)>::_FUN(void *) () at /usr/src/debug/wpebackend-fdo/1.6.1-r0/wpebackend-fdo-1.6.1/src/view-backend-exportable-private.h:124
#9  0x73ec3f38 in wpe_view_backend_destroy (backend=0x6f0b1bd8) at /usr/src/debug/libwpe/1.6.0-r0/libwpe-1.6.0/src/view-backend.c:59
#10 0x6fb11e60 in wpe_view_backend_exportable_fdo_destroy (exportable=0x6f0b18e8) at /usr/src/debug/wpebackend-fdo/1.6.1-r0/wpebackend-fdo-1.6.1/src/view-backend-exportable-fdo.cpp:182
#11 0x74714518 in void WTF::derefGPtr<_WebKitWebViewBackend>(_WebKitWebViewBackend*) () from /opt/veld/1.0.0/sysroots/cortexa9hf-neon-veld-linux-gnueabi/usr/lib/libWPEWebKit-1.0.so.3.10.6
#12 0x7470d6e6 in _WebKitWebViewPrivate::~_WebKitWebViewPrivate() () from /opt/veld/1.0.0/sysroots/cortexa9hf-neon-veld-linux-gnueabi/usr/lib/libWPEWebKit-1.0.so.3.10.6
#13 0x7470bddc in webkit_web_view_finalize(_GObject*) () from /opt/veld/1.0.0/sysroots/cortexa9hf-neon-veld-linux-gnueabi/usr/lib/libWPEWebKit-1.0.so.3.10.6
#14 0x74009df4 in g_object_unref (_object=0x6f0b5080) at ../glib-2.58.3/gobject/gobject.c:3346
#15 0x763e8240 in cog_shell_remove_view (shell=shell@entry=0x6f007de0, name=<optimized out>) at /usr/src/debug/cog/git-r0/git/core/cog-shell.c:526
```
The client itself in frame 2 looks like an already freed chunk of memory:

```
(gdb) print *client
$23 = {connection = 0x9, source = 0xa, display = 0x25, display_resource = 0x0, id_count = 0, mask = 36, link = {prev = 0x0, next = 0x0}, objects = {client_entries = {size = 18, alloc = 50, data = 0x0}, 
    server_entries = {size = 0, alloc = 0, data = 0x0}, side = 0, free_list = 0}, destroy_signal = {listener_list = {prev = 0x0, next = 0x0}, emit_list = {prev = 0x0, next = 0x0}}, ucred = {pid = 0, uid = 0, 
    gid = 0}, error = 33, resource_created_signal = {listener_list = {prev = 0x6f000044, next = 0x6f0740a8}, emit_list = {prev = 0x6f00b548, next = 0x6f00b548}}}
```

psaavedra: That line of code was added apparently to fix some crash that it was happening at some point in the "Test code" attached in 
https://bugs.webkit.org/show_bug.cgi?id=218612. I can not reproduce anymore this crash even revert this commit so probably it is a superfluous change actually not needed.

aperezdc: Should we revert the changes from this PR, then?

psaavedra: Reverting 9a0e6cb62bfae44ef3496d950899ecb34ddd24a5 I found this new crash:

```
#1  0x703c0b04 in wl_priv_signal_emit (signal=signal@entry=0x6f0f6960, data=data@entry=0x6f0f6920) at /usr/src/debug/wayland/1.13.0-r0/wayland-1.13.0/src/wayland-server.c:2008
#2  0x703c0c70 in wl_client_destroy (client=client@entry=0x6f0f6920) at /usr/src/debug/wayland/1.13.0-r0/wayland-1.13.0/src/wayland-server.c:852
#3  0x703c0d4c in wl_client_connection_data (fd=<optimized out>, mask=<optimized out>, data=0x6f0f6920) at /usr/src/debug/wayland/1.13.0-r0/wayland-1.13.0/src/wayland-server.c:341
#4  0x703c2a7c in wl_event_loop_dispatch (loop=0x6f00f818, timeout=timeout@entry=0) at /usr/src/debug/wayland/1.13.0-r0/wayland-1.13.0/src/event-loop.c:423
>>>>   #5  0x6fb1a758 in WS::ServerSource::<lambda(GSource*, GSourceFunc, gpointer)>::operator() (__closure=0x0, base=0x6f028780) at /usr/src/debug/wpebackend-fdo/1.6.1-r0/wpebackend-fdo-1.6.1/src/ws.cpp:120
#6  WS::ServerSource::<lambda(GSource*, GSourceFunc, gpointer)>::_FUN(GSource *, GSourceFunc, gpointer) () at /usr/src/debug/wpebackend-fdo/1.6.1-r0/wpebackend-fdo-1.6.1/src/ws.cpp:129
#7  0x73f2f668 in g_main_dispatch (context=0x1402740) at ../glib-2.58.3/glib/gmain.c:3182
#8  g_main_context_dispatch (context=context@entry=0x1402740) at ../glib-2.58.3/glib/gmain.c:3874
#9  0x73f2f88c in g_main_context_iterate (context=0x1402740, block=block@entry=1, dispatch=dispatch@entry=1, self=<optimized out>) at ../glib-2.58.3/glib/gmain.c:3947
#10 0x73f2fc2c in g_main_loop_run (loop=0x13e42c0) at ../glib-2.58.3/glib/gmain.c:4143
#11 0x004aea70 in wpe::WpeThreadedLauncher::runThread (this=this@entry=0x1404de8) at /usr/src/debug/annonymous-wpe-launcher/999-r0/git/src/WpeThreadedLauncher.cc:1005
#12 0x004ba45c in wpe::WpeThreadedLauncher::<lambda(void*)>::operator() (__closure=0x0, data=<optimized out>) at /usr/src/debug/annonymous-wpe-launcher/999-r0/git/src/WpeThreadedLauncher.cc:585
#13 wpe::WpeThreadedLauncher::<lambda(void*)>::_FUN(void *) () at /usr/src/debug/annonymous-wpe-launcher/999-r0/git/src/WpeThreadedLauncher.cc:589
#14 0x73f5bbdc in g_thread_proxy (data=0x140d000) at ../glib-2.58.3/glib/gthread.c:784
#15 0x76624e8c in start_thread (arg=0x6f9ff2b0) at /usr/src/debug/glibc/2.26-r0/git/nptl/pthread_create.c:465
#16 0x73cc7448 in ?? () at ../sysdeps/unix/sysv/linux/arm/clone.S:73 from /mnt/data-external/image/sysroots/cortexa9hf-neon-image-linux-gnueabi/lib/libc-2.26.so
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```



Both crashes are happening in the `wl_client_destroy()`. The one that you found is in the `WS::Instance::unregisterViewBackend`,  the new one it is happening in the `wl_event_loop_dispatch()`. Both in the `wayland-server.c`  

AFAIK the UI-process is acting as Wayland server for the WebProcess so what is happending in both cases is, the wl_client_destroy() invoked for a wl_client already closed. I think, I saw it in the code in the past, the wl_client is closed in by wayland if there is a error in the communication (ex: the FD is closed because the other peer of the waylannd channel is gone, aka: WebProcess dead). Probably what we are missing a safe-guard (like `if (m_client.object)`) in some place.
    

The worklow looks like the following:


1. The WebProcess terminates and (as ws-client) sends an UnregisterSurface IPC message:

    ```
   usr/src/debug/wpebackend-fdo/1.6.1-r0/wpebackend-fdo-1.6.1/src/ws-client.cpp:        m_glib.socket->send(FdoIPC::Messages::UnregisterSurface, m_wl.wpeBridgeId);
    ```

2. The UI-process handles the message:

``` cpp
void ViewBackend::didReceiveMessage(uint32_t messageId, uint32_t messageBody)
{
    switch (messageId) {
    case FdoIPC::Messages::RegisterSurface:
        registerSurface(messageBody);
        break;
    case FdoIPC::Messages::UnregisterSurface:
        unregisterSurface(messageBody); <<<<<<<<<<
        break;
    default:
        assert(!"WPE fdo received an invalid IPC message");
    }
}
```

3. This also means: `ViewBackend::unregisterSurface()` -> `Instance::unregisterViewBackend()`


   (maybe the wl_client should be in the `ViewBackend::unregisterSurface()` and not in the `Instance`, because it is this first one who knows the existence of this wl_client).

3. a) (**with** 9a0e6cb62bfae44ef3496d950899ecb34ddd24a5) The `wl_client_destroy(it->second->client)` the idea was to prevent this client been called in the next loop iteration of the WS GSource  (`m_source`):
    ```
   #7  0x6fb15e74 in <lambda(void*)>::operator() (__closure=<optimized out>, data=<optimized out>) at /usr/src/debug/wpebackend-fdo/1.6.1-r0/wpebackend-fdo-1.6.1/src/view-backend-exportable-private.h:123
   ```

   ``` cpp
    // dispatch
    [](GSource* base, GSourceFunc, gpointer) -> gboolean                                                               
    {                                                                                                                  
        auto& source = *reinterpret_cast<ServerSource*>(base);                                                         
                                                                                                                       
        if (source.pfd.revents & G_IO_IN) {
            struct wl_event_loop* eventLoop = wl_display_get_event_loop(source.display);                               
            wl_event_loop_dispatch(eventLoop, 0); <<<<<<<<<<<<<<<<<<
            wl_display_flush_clients(source.display);                                                                  
        }
                                                                                                                       
        if (source.pfd.revents & (G_IO_ERR | G_IO_HUP))                                                                
            return FALSE;

        source.pfd.revents = 0;
        return TRUE;
    },
    ```

    but in your case this `wl_client_destroy()` call comes too late leading in a double destroy action.

3. b) (**without** 9a0e6cb62bfae44ef3496d950899ecb34ddd24a5), the GSource loop is dispatching events to a wl_client what it was already freed but still being referenciated for someone else (the queued GSource event?) so we are accessing to a memory what it could be free or not and in any case not valid anymore.

The solution could be something like safe-guard the `wl_event_loop_dispatch()` with something like `if (m_client.object)` but the dispatch events for all the backends registered in the WS::Instance so ...

PR #128, opened 2020-11-18
meson: Only add a wayland-egl dependency if needed

aperezdc:
Some drivers provide the `wl_egl_*` symbols their EGL library (either directly or indirectly in some other one needed by it) and do not ship a `wayland-egl` library. In that case, checking for `wayland-egl` fails even if the symbols are usable. To accomodate for this, check whether `wl_egl_create_window` is usable before trying to add the dependency on `wayland-egl`.

----

Idea from https://github.com/Igalia/cog/commit/170e3ef9d and also a patch proposed by @psaavedra (thanks!)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -244,9 +244,40 @@ deps = [
 	dependency('gobject-2.0'),
 	dependency('wayland-client'),
 	dependency('wayland-server'),
-	dependency('wayland-egl'),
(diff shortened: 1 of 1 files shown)


psaavedra (approved):
LGTM.

aperezdc:
@psaavedra Could you please review again? I have changed the logic to make sure it will work when `libwpe` is built as a subproject, and now the CI passes :ok_hand:

psaavedra (approved):

PR #130, opened 2020-11-24
Add unstable dmabuf-pool API

zdobersek:
Introduce wpe_view_backend_dmabuf_pool_fdo, a wpe_view_backend extension that
demands manual dmabuf managemenet from the user.

This approach to buffer management allows for most flexibility in terms of what
kind of dmabuf objects should be used for rendering. It allows for user to
specify in detail the format, multi-planarity and modifier data of the dmabuf
object.

In order to work with such dmabuf objects, the renderer_backend_egl interfaces
have to leverage a new WS::EGLClient implementation that uses custom Wayland
protocols to properly handle dmabuf content across process boundaries. To
correctly render into these objects, we have to leverage the surfaceless EGL
platform, along with importing the dmabuf objects into EGLImages that we then
use as backing resources for renderbuffers used to render a given frame.

For each wpe_view_backend_dmabuf_pool_fdo, the related client (managed by the
user) has to provide callback implementations that handle dmabuf pool entry
management as well as the notification of which dmabuf pool entry has been
committed for display as the content of the related view.

Upon the create_entry callback dispatch on the client, the callback (as provided
by the user) has to return a wpe_dmabuf_pool_entry object that was initialized
with the relevant dmabuf data. Upon the destroy_entry callback dispatch, the
specified object has to be destroyed.

The commit_entry callback dispatch on the client specifies which dmabuf object
should be presented on the screen, containing the most-recent composition of the
corresponding view's content.

The dmabuf pool entry allocations and deallocations are relayed from the
renderer_backend_target_egl implementation. Whenever a new buffer is required
for rendering, the renderer_backend_target_egl will request for a pool entry to
be created, getting back the relevant dmabuf information and spawning necessary
GL resources that are needed to render into that dmabuf. When finished, the
pool entry will be committed, leaving it up to the user to handle the pool entry
as required.

TODO:
  - TargetDmabufPool: cleanup
  - TargetDmabufPool: resizing support

### DIFF
--- a/src/egl-client-dmabuf-pool.cpp
+++ b/src/egl-client-dmabuf-pool.cpp
@@ -1,0 +1,369 @@
+/*
+ * Copyright (C) 2020 Igalia S.L.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+ * PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "egl-client-dmabuf-pool.h"
+
+#include "ws-client.h"
+
+#include <array>
+#include <cstdio>
+
+namespace WS {
+namespace EGLClient {
+
+BackendDmabufPool::BackendDmabufPool(BaseBackend&)
+{
+}
+
+BackendDmabufPool::~BackendDmabufPool() = default;
+
+EGLNativeDisplayType BackendDmabufPool::nativeDisplay() const
+{
+    return EGL_DEFAULT_DISPLAY;
+}
+
+uint32_t BackendDmabufPool::platform() const
+{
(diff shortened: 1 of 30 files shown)


src/egl-client-dmabuf-pool.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+                std::copy(planeAttributes.begin(), planeAttributes.end(),
+                    std::next(attributes.begin(), attributesCount));
+                attributesCount += planeAttributes.size();
+            }
+            attributes[attributesCount++] = EGL_NONE;
+        }
+
+        buffer->egl.image = m_renderer.createImageKHR(eglGetCurrentDisplay(),
+            EGL_NO_CONTEXT, EGL_LINUX_DMA_BUF_EXT, nullptr, attributes.data());
```
aperezdc, line 235: I am missing here checking for errors returned by `eglCreateImage[KHR]`. While I know that it is hardly expected to fail because the parameters passed are under our control here and expected to be valid, it's reasonable to think that at least `EGL_BAD_ALLOC` can be produced (when out of memory)—and it that case it would be nice to at least have something logged out.

The same applies to some of the rest of the GL code around.

src/view-backend-exportable-fdo-egl.cpp
```diff
@@ -118,6 +118,17 @@ class ClientBundleEGLDeprecated final : public ClientBundle {
         assert(!"should not be reached");
     }
 
+    struct wpe_dmabuf_pool_entry* createDmabufPoolEntry() override
+    {
+        assert(!"should not be reached");
```
aperezdc, line 123: I would go for `g_assert_not_reached()` here and in the instances below, WDYT?

```suggestion
        g_assert_not_reached();
```

aperezdc (changes requested):
I've had this half-reviewed in a tab for weeks, sorry about the delay!

Functionality-wise patch looks fine to me, with a few nits. Something I would like to start doing is documentation with every commit that adds API. We have scaffolding in place for that, and in this particular patch we could really use a copy of the buffer lifetime notes from the commit log saved to a Markdown file e.g. `docs/dmabufpool.markdown`, and that file listed in `docs/sitemap.txt`.

aperezdc:
This PR needs a rebase. Do we still want to try and merge this before branching for 1.10.x?

src/view-backend-exportable-fdo-egl.cpp
zdobersek: This is aligned with what the other dead-end overrides do.

src/egl-client-dmabuf-pool.cpp
zdobersek: Returned EGLImageKHR deserves a null-check.

Best I can do for GL code is framebuffer-completeness check at the end of this function.

aperezdc (approved):

src/renderer-backend-egl.cpp
```diff
@@ -170,6 +177,14 @@ struct wpe_renderer_backend_egl_target_interface fdo_renderer_backend_egl_target
...
         target.m_impl->frameRendered();
     },
+#if WPE_CHECK_VERSION(1,9,1)
+    // deinitialize
+    [](void* data)
+    {
+        auto& target = *reinterpret_cast<Target*>(data);
+        target.m_impl->deinitialize();
+    },
+#endif
```
aperezdc, line 187: The `WPE_CHECK_VERSION` conditional here means that when building with an older libwpe the `::deinitialize()` method will never get called, and we would leak resources.

Wouldn't it make sense to only compile the whole DMA-BUF pool support when building againsy  libwpe≥1.9.1?

src/egl-client-dmabuf-pool.cpp
```diff
@@ -0,0 +1,317 @@
...
+{
+    wl_list_init(&m_buffer.list);
+
+    m_renderer.width = width;
+    m_renderer.height = height;
+}
+
+TargetDmabufPool::~TargetDmabufPool()
+{
+    wl_list_init(&m_buffer.list);
```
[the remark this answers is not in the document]
aperezdc: Okay, I see that the list is emptied in the `::deinitialize()` method, that's alright.

aperezdc (commented):
@zdobersek, I have one last doubt before merging this, please check my comment about the version check for libwpe≥1.9.1

aperezdc:
After a chat with @zdobersek today we have agreed to slate this for 1.12.x, so I have changed the milestone accordingly.

PR #138, opened 2021-03-03
exported-buffer-shm.h: add a wl_resource getter

zdobersek:
Add the wpe_fdo_shm_exported_buffer_get_resource() entrypoint,
allowing the recepient of the export to also access the wl_resource
object and using that to e.g. attach destroy listeners to that
resource object.

### DIFF
--- a/include/wpe/exported-buffer-shm.h
+++ b/include/wpe/exported-buffer-shm.h
@@ -35,7 +35,11 @@ extern "C" {
 #endif
 
 struct wpe_fdo_shm_exported_buffer;
+struct wl_resource;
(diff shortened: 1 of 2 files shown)


psaavedra (approved):
Just for curiosity, I see this getter convinient in general but is there any particular case where this can be useful rigth now?

zdobersek:
I use it in https://github.com/Igalia/cog/pull/267 for caching buffer content based on the exported resource. `wl_resource` seems the best choice for that since it's persistent across different frames (as long as the client is able to reuse it).

aperezdc (approved):

PR #139, opened 2021-03-03
Avoid generated version header

aperezdc:
Use a fixed `wpebackend-fdo-version.h` header which contains `#defines` for the version numbers only, which is the location where the canonical version is read by the build system, and also used by `version.h`, which does not need to be generated anymore.

This avoids needing to use relative `#include` paths, which continued being a pitfall even after the recent improvements due to the mixup of headers with the same name (and relative path!) in the preprocessor search path e.g. `wpe/version.h` from libwpe.

----

This follows the same principle as https://github.com/WebPlatformForEmbedded/libwpe/pull/76

### DIFF
--- a/include/wpe/version.h
+++ b/include/wpe/version.h
@@ -30,6 +30,8 @@
 #ifndef __wpe_fdo_version_h__
 #define __wpe_fdo_version_h__
 
+#include "wpebackend-fdo-version.h"
+
 /**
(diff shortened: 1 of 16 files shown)


carlosgcampos (approved):

aperezdc:
Thanks @carlosgcampos :smile_cat:

PR #144, opened 2021-03-29
view backend: Properly unregister surfaces

aperezdc:
When a new surface is registered for an existing `ViewBackend`, unregister the old one first, to avoid keeping stale registrations. This makes the `WS::Instance::m_viewBackendMap` only contain one entry pointing for any given `ViewBackend` instance, instead of growing continuously.

To ensure that the same code path is taken when the `wl_client` for a `ViewBackend` is destroyed, all the operations to be done are moved into `ViewBackend::clientDestroyNotify()`, the callback of the `wl_client` destruction listener, then `ViewBackend::unregisterSurface()` only needs to call `wl_client_destroy()`. This way the `~ViewBackend` destructor only needs to call `unregisterSurface()` as well.

Also, avoid doing any operation in `ViewBackend::dispatchFrameCallbacks()` and `ViewBackend::releaseBuffer()` if the client connection has been lost.

### DIFF
--- a/src/view-backend-private.cpp
+++ b/src/view-backend-private.cpp
@@ -36,14 +36,12 @@ ViewBackend::ViewBackend(ClientBundle* clientBundle, struct wpe_view_backend* ba
     m_clientBundle->viewBackend = this;
 
     wl_list_init(&m_frameCallbacks);
+    wl_list_init(&m_clientDestroy.link);
 }
 
 ViewBackend::~ViewBackend()
 {
-    clearFrameCallbacks();
-
     unregisterSurface(m_surfaceId);
-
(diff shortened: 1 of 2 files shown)


aperezdc:
The most consistent way to reproduce the issue fixed by this PR is to use Cog and trigger PSON. What I have been doing is:

1. Run `cog --platform=fdo ddg.gg`.
2. Search for “webgl demos”.
3. Choose the result that navigates to https://webglsamples.org — for me it's the first one.

Without this fix, the load in step (3.) will get stuck.

zdobersek (approved):

aperezdc:
Thanks for taking a look, @zdobersek 🤗

PR #147, opened 2021-04-08
docs: Add architecture section, describe nested compositor

aperezdc:
This adds a “rough strokes” view of the nested compositor.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,0 +1,40 @@
+# Architecture
+
+## Nested Compositor
+
+The FDO backend exposes a minimal Wayland compositor to the `WPEWebProcess`
+instances which take care of the actual rendering of web content. The Wayland
+protocol is thus used between the “UI process” (your application) and the
+processes that take care of rendering—Wayland is, after all, an IPC protocol
+designed to pass graphics buffers around.
+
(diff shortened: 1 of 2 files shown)


mcatanzaro:
I think it would be useful to add a comment on the trust boundary between the unsandboxed UI process code, e.g. ws.cpp, which is not allowed to trust the data it receives from the sandboxed web process code, e.g. ws-client.cpp. It's tempting to think of the IPC peer as trusted, since the code is wpebackend-fdo on both sides of the IPC, but one side is the WebKit web process, after all, and the other side is the UI process. The UI process is supposed to be robust to crafted malicious IPC sent by a compromised malicious web process: if input is not carefully validated, we could have a sandbox escape.

For example, in #148 I added a request that ws-client.cpp uses to tell ws.cpp to free something. I woke up today and thought "is that really safe if the web process has been compromised by malicious web content? What if the message is sent multiple times, could that trigger a use-after-free in the UI process?" In this particular case, it really is safe, because a side effect causes the IPC to be disconnected, but I hadn't even considered that possibility during development.

aperezdc:
The description here is not quite right, so I will close this for now. We will still want to write down some documentation at some point, but I do not have the bandwidth at the moment.

PR #148, opened 2021-04-09
Stop crashing!

### LINKED ISSUE #145: crash

mcatanzaro:
Currently, we crash if process-swapping back to a process that
previously had an accelerated surface. The crash occurs because the web
process is not prepared for its connection to the UI process to fail.
The UI process recently began closing its connection to the web process
too soon. It incorrectly assumed that it's safe to close the connection
when a surface is deregistered, which corresponds to the destruction of
a WS::BaseTarget in the web process. But in fact, the connection to the
web process needs to be kept open until the WS::BaseBackend is destroyed
in the web process. Each BaseBackend may have more than one BaseTarget.
A new BaseTarget is created when displaying a new page.

Reverting this sequence of commits avoids this misunderstanding and fixes #145.

Finally, I add some asserts to ensure we crash ASAP if globals are not properly bound in the future.

### DIFF
--- a/src/ws-client.cpp
+++ b/src/ws-client.cpp
@@ -126,6 +126,9 @@ BaseBackend::BaseBackend(int hostFD)
     wl_display_roundtrip(m_wl.display);
     wl_registry_destroy(registry);
 
+    if (!m_wl.wpeBridge)
+        g_error("Failed to bind wpe_bridge");
+
     wpe_bridge_add_listener(m_wl.wpeBridge, &s_bridgeListener, this);
     wpe_bridge_initialize(m_wl.wpeBridge);
     wl_display_roundtrip(m_wl.display);
@@ -203,6 +206,11 @@ void BaseTarget::initialize(BaseBackend& backend)
     wl_registry_add_listener(registry, &s_registryListener, this);
     wl_display_roundtrip_queue(display, m_wl.eventQueue);
     wl_registry_destroy(registry);
+
+    if (!m_wl.compositor)
+        g_error("Failed to bind wl_compositor");
+    if (!m_wl.wpeBridge)
+        g_error("Failed to bind wpe_bridge");
 
     m_wl.surface = wl_compositor_create_surface(m_wl.compositor);
     wl_proxy_set_queue(reinterpret_cast<struct wl_proxy*>(m_wl.surface), m_wl.eventQueue);
--- a/src/ws.cpp
+++ b/src/ws.cpp
@@ -532,6 +532,7 @@ void Instance::unregisterViewBackend(uint32_t surfaceId)
     auto it = m_viewBackendMap.find(surfaceId);
     if (it != m_viewBackendMap.end()) {
         it->second->apiClient = nullptr;
+        wl_client_destroy(it->second->client);
         m_viewBackendMap.erase(it);
     }
 }
--- a/src/ws.cpp
+++ b/src/ws.cpp
@@ -532,7 +532,6 @@ void Instance::unregisterViewBackend(uint32_t surfaceId)
     auto it = m_viewBackendMap.find(surfaceId);
     if (it != m_viewBackendMap.end()) {
         it->second->apiClient = nullptr;
-        wl_client_destroy(it->second->client);
         m_viewBackendMap.erase(it);
     }
 }
--- a/src/extensions/audio.cpp
+++ b/src/extensions/audio.cpp
@@ -159,6 +159,9 @@ public:
         wl_display_roundtrip_queue(display, eventQueue);
         wl_registry_destroy(registry);
 
+        if (!m_wl.audio)
+            g_error("Failed to bind wpe_audio");
+
         wl_event_queue_destroy(eventQueue);
     }
 
--- a/src/extensions/video-plane-display-dmabuf.cpp
+++ b/src/extensions/video-plane-display-dmabuf.cpp
@@ -159,6 +159,9 @@ public:
         wl_display_roundtrip_queue(display, eventQueue);
         wl_registry_destroy(registry);
 
+        if (!m_wl.videoPlaneDisplayDmaBuf)
+            g_error("Failed to bind wpe_video_plane_display_dmabuf");
+
         wl_event_queue_destroy(eventQueue);
     }
 
--- a/src/view-backend-private.cpp
+++ b/src/view-backend-private.cpp
@@ -139,8 +139,6 @@ void ViewBackend::unregisterSurface(uint32_t surfaceId)
 
     clearFrameCallbacks();
 
-    g_clear_pointer(&m_client.object, wl_client_destroy);
-
     WS::Instance::singleton().unregisterViewBackend(m_surfaceId);
     m_surfaceId = 0;
 }
--- a/src/view-backend-private.cpp
+++ b/src/view-backend-private.cpp
@@ -36,12 +36,14 @@ ViewBackend::ViewBackend(ClientBundle* clientBundle, struct wpe_view_backend* ba
     m_clientBundle->viewBackend = this;
 
     wl_list_init(&m_frameCallbacks);
-    wl_list_init(&m_clientDestroy.link);
 }
 
 ViewBackend::~ViewBackend()
 {
+    clearFrameCallbacks();
+
     unregisterSurface(m_surfaceId);
+
     if (m_clientFd != -1)
         close(m_clientFd);
 }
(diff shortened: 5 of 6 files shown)


src/ws.cpp
```diff
@@ -187,6 +187,11 @@ static const struct wpe_bridge_interface s_wpeBridgeInterface = {
         wpe_bridge_send_connected(resource, bridgeID);
         Instance::singleton().registerSurface(bridgeID, surface);
     },
+    // disconnect
```
psaavedra, line 190: This disconnect message to make an explicit wl_client_destroy makes and could to be the definitive solution to avoid crashes related with wl_client structuctures being  destroyed in `wl_client_connection_data()` (check explanation in https://github.com/Igalia/WPEBackend-fdo/pull/140) because a wayland error (fd/socket ERROR) for example, because the client is gone.

Looks good to me.

psaavedra (commented):
We are sure this commit is not  introducing crashes? 

OTOH, I've noticed a change of behavior after this commit using the cog's multiple-views branch. With this patch, I'm observing situations where the surface related to the second view  (previously not active) keeps not refreshed after make it active again. //CC @aperezdc

psaavedra (commented):
OK reverting this because the destruction is properly triggered by the `disconnect` message https://github.com/Igalia/WPEBackend-fdo/pull/148/commits/8f88d1a468e581cf7a7f2483809e49c3cd82508f.

psaavedra (commented):
in reference to c93d1e9a16a16f001efade793833ca5c85d0f3a3. DITTO. OK reverting this because the destruction is properly triggered by the `disconnect` message https://github.com/Igalia/WPEBackend-fdo/pull/148/commits/8f88d1a468e581cf7a7f2483809e49c3cd82508f.

mcatanzaro:
> We are sure this commit is introducing crashes?

Hm, not on its own, but it destroys the wl_client in ViewBackend::unregisterSurface, which needs to not happen. I haven't looked closely enough to know if that can be disentangled from the rest of the commit....

> OTOH, I've noticed a change of behavior after this commit using the cog's multiple-views branch. With this patch, I'm observing situations where the surface related to the second view (previously not active) keeps not refreshed after make it active again. //CC @aperezdc

I haven't noticed this in Ephy, but I'm not surprised. :/

aperezdc:
@mcatanzaro As you can imagine from our chats the past days, I think we are on a good path with this—I need to do some testing with this using Cog's `devel/multiple-views` branch today.

mcatanzaro:
> Hm, not on its own, but it destroys the wl_client in ViewBackend::unregisterSurface, which needs to not happen. I haven't looked closely enough to know if that can be disentangled from the rest of the commit....

I would need some help from Adrian to avoid reverting that commit. Problem if I do not revert 99bd04019800f84a722ae99bf1a352f225d93002 is ViewBackend::unregisterSurface would no longer do anything at all. Currently all it does is call wl_client_destroy(), but that's too soon to call wl_client_destroy(). The actual surface unregistration happens in ViewBackend::clientDestroyNotify, but that's not a good place because the lifetime of the surface (corresponding to a BaseTarget) is shorter than the lifetime of the client (corresponding to a BaseBackend). So yes, that commit really needs to be reverted IMO, and reworked with this in mind.

src/ws-client.cpp
```diff
@@ -204,6 +208,11 @@ void BaseTarget::initialize(BaseBackend& backend)
     wl_display_roundtrip_queue(display, m_wl.eventQueue);
     wl_registry_destroy(registry);
 
+    if (!m_wl.compositor)
+        g_error("Failed to bind wl_compositor");
+    if (!m_wl.wpe_bridge)
```
aperezdc, line 213: This should be `wpeBridge`, the compiler complains:

```
../src/ws-client.cpp: In member function ‘void WS::BaseTarget::initialize(WS::BaseBackend&)’:
../src/ws-client.cpp:213:15: error: ‘struct WS::BaseTarget::<unnamed>’ has no member named ‘wpe_bridge’; did you mean ‘wpeBridge’?
  213 |     if (!m_wl.wpe_bridge)
      |               ^~~~~~~~~~
      |               wpeBridge
```

aperezdc: (On a related note, the CI somehow did not run for this PR filed from your fork, I have not seen that happen before, but maybe we have something misconfigured.)

aperezdc (changes requested):

mcatanzaro:
> Call wl_client_destroy() when the client is done with us
> 
> Currently, we crash if process-swapping back to a process that
> previously had an accelerated surface. The crash occurs because the web
> process is not prepared for its connection to the UI process to fail.
> The UI process recently began closing its connection to the web process
> too soon. It incorrectly assumed that it's safe to close the connection
> when a surface is deregistered, which corresponds to the destruction of
> a WS::BaseTarget in the web process. But in fact, the connection to the
> web process needs to be kept open until the WS::BaseBackend is destroyed
> in the web process. Each BaseBackend may have more than one BaseTarget.
> A new BaseTarget is created when displaying a new page.
> 
> So instead, let's not destroy the client until the BaseBackend is
> destroyed. We signal this by creating a wpe_bridge_disconnect() request
> to signal the UI process that we are done with it.
> 
> This fixes #145, sort of. The crash was actually fixed by the previous
> sequence of reverts. But after those reverts, the wl_client object would
> be leaked in the UI process until the UI process quits. This commit
> avoids that leak.

OK, Adrian noticed this doesn't actually work. The server never receives the wl_client_destroy() message. I tried adding a roundtrip after the call to wpe_bridge_disconnect(), but that didn't work either.

I'm going to remove this broken commit from the PR for now, and edit the first comment of this PR accordingly. This means we might leak the wl_client on the server side, although Adrian isn't certain about that, as libwayland itself frees it in various codepaths.

mcatanzaro:
> OK, Adrian noticed this doesn't actually work. The server never receives the wl_client_destroy() message. I tried adding a roundtrip after the call to wpe_bridge_disconnect(), but that didn't work either.
> 
> I'm going to remove this broken commit from the PR for now, and edit the first comment of this PR accordingly. This means we might leak the wl_client on the server side, although Adrian isn't certain about that, as libwayland itself frees it in various codepaths.

Adrian noticed that `~BaseBackend` is never called. The BaseBackend is leaked when the web process terminates!

mcatanzaro:
> OTOH, I've noticed a change of behavior after this commit using the cog's multiple-views branch. With this patch, I'm observing situations where the surface related to the second view (previously not active) keeps not refreshed after make it active again. //CC @aperezdc

Adrian thinks we need to revert more commits. He's working on a different branch.

The only remaining useful thing is this MR is my asserts. I will open a new MR for those. Closing this.

mcatanzaro:
> Adrian noticed that ~BaseBackend is never called. The BaseBackend is leaked when the web process terminates!

This happens because wpe_renderer_backend_egl_destroy() is never called, for different reasons for WPE and GTK port. For WPE, Adrian says some exit handlers that destroy PlatformDisplayLibWPE are mysteriously not running. For GTK, it's because the PlatformDisplayLibWPE is owned by the WebProcess object, which is never destroyed. I tried to fix it by destroying the PlatformDisplayLibWPE in WebProcess::platformTerminate:

```
void WebProcess::platformTerminate()
{
#if PLATFORM(WAYLAND) && USE(WPE_RENDERER)
    m_wpeDisplay = nullptr;
#endif
}
```

But turns out nothing ever calls WebProcess::terminate....

aperezdc:
> > OK, Adrian noticed this doesn't actually work. The server never receives the wl_client_destroy() message. I tried adding a roundtrip after the call to wpe_bridge_disconnect(), but that didn't work either.
> > I'm going to remove this broken commit from the PR for now, and edit the first comment of this PR accordingly. This means we might leak the wl_client on the server side, although Adrian isn't certain about that, as libwayland itself frees it in various codepaths.
> 
> Adrian noticed that `~BaseBackend` is never called. The BaseBackend is leaked when the web process terminates!



> > Adrian noticed that ~BaseBackend is never called. The BaseBackend is leaked when the web process terminates!
> 
> This happens because wpe_renderer_backend_egl_destroy() is never called, for different reasons for WPE and GTK port. For WPE, Adrian says some exit handlers that destroy PlatformDisplayLibWPE are mysteriously not running.

This would be [WebKit bug #217655](https://bugs.webkit.org/show_bug.cgi?id=217655], which we haven't had the time to investigate properly yet. Even if this does not get to run, the Wayland client connection that the web process had opened to the nested compositor will nevertheless get closed and the backend should handle dropped connections gracefully (which is part of what the whole wild goose chase has been about).

> For GTK, it's because the PlatformDisplayLibWPE is owned by the WebProcess object, which is never destroyed. I tried to fix it by destroying the PlatformDisplayLibWPE in WebProcess::platformTerminate:
> 
> ```
> void WebProcess::platformTerminate()
> {
> #if PLATFORM(WAYLAND) && USE(WPE_RENDERER)
>     m_wpeDisplay = nullptr;
> #endif
> }
> ```
> 
> But turns out nothing ever calls WebProcess::terminate....

- In `Source/WebKit/WebProcess/WebProcess.cpp` function `WebProcess::terminate()` there is a call to `platformTerminate()`.
- The `WebProcess::terminate()` method overrides `AuxiliaryProcess::terminate()`
- Both `AuxiliaryProcess::shutDown()` and `AuxiliaryProcess::terminationTimerFired()` call `terminate()`.
  - The only way (I think) in which `::shutDown()` gets called is by sending a `ShutDown` message (see `AuxiliaryProcess.messages.in`!)
- The `m_terminationTimer` is controlled by `::enableTermination()` and `::disableTermination()`.
  - For WPE there are calls to those only at `::createWebPage()`, which calls `::disableTermination()`; and `::removeWebPage()` which calls `::enableTermination()—this makes sense, the process is to be kept alive while there is at least one page being handled by it.

So the question that remains is: is there any other case (for example: abrupt termination) in which `::terminate()` should be called? Otherwise the above, while convoluted, seems okay to me.

mcatanzaro:
In GTK port it seems that `WebProcess::terminate` is never called. At least, my printf does not print. ;)

PR #150, opened 2021-04-10
Add some checks to ensure we properly bind server globals

mcatanzaro:
If we fail to bind any global objects, we are going to crash. Let's do
that as nicely as possible using g_error(), rather than dereferencing
null pointers later on.

Suggested by #145


aperezdc (approved):

aperezdc:
Thanks @mcatanzaro 😃

PR #151, opened 2021-04-12
A couple of wpe_bridge related cleanups

aperezdc:
This has two commits with changes that (IMO) make the code a tad easier to follow:

* The `Surface::id` and `Surface::client` members can both be replaced by keeping around the  `wl_resource`, and query its client with `wl_resource_get_client()`; the identifier itself is never used—and if ever needed, there is `wl_resource_get_id()` which could be used.
* The `WS::Singleton::m_viewBackendMap` associated an identifier generated by the `wpe_bridge` interface with a given `Surface`, hence using “surfaceId” around is misleading and changing variables to use “bridgeId” seems more appropriate.

### DIFF
--- a/src/view-backend-private.cpp
+++ b/src/view-backend-private.cpp
@@ -41,7 +41,7 @@ ViewBackend::ViewBackend(ClientBundle* clientBundle, struct wpe_view_backend* ba
 
 ViewBackend::~ViewBackend()
 {
-    unregisterSurface(m_surfaceId);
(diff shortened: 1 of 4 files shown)


zdobersek (approved):

aperezdc:
Cherry-picked into the `wpebackend-fdo-1.8` branch as commits 496b5b0e16b5aedcebefb7d668f8476bbefbe648 and 2809478c9725636e6ea680f241810dc7e59877bb.

PR #152, opened 2021-04-14
Meson: Fix the build with Python 3.5

aperezdc:
In theory we support building with Python as old as 3.5, but the CI was using 3.6. This downgrades the version used when doing test builds, and fixes an issue in the `scripts/version.py` helper script.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v1
         with:
-          python-version: 3.6
+          python-version: 3.5
       - name: Python Package Cache
         uses: actions/cache@v1
         with:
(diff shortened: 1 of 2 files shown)


aperezdc:
This will need a similar fix in `libwpe`, which is the reason why it fails to build as subproject.

psaavedra (approved):
For the record. This fix solves this error on Python 3.5:

```
Traceback (most recent call last):
  File "./scripts/version.py", line 21, in <module>
    version[m[1]] = m[2]
TypeError: '_sre.SRE_Match' object is not subscriptable
```

aperezdc:
Cherry-picked into the `wpebackend-fdo-1.10` branch as commits 076a92d2ce1a6f929328db3699f189743439d09c and 89e4b1683ad51c1b066fc10c748e5d524fb37cd2.

PR #153, opened 2021-04-14
ws: Remove wl_client pointer that can end up dangling

### LINKED ISSUE #145: crash

aperezdc:
Instead of trying to keep track of the destruction of the `wl_client` referenced by `ViewBackend::m_client`, completely remove the member variable and instead:
    
- In `ViewBackend::releaseBuffer()` use `wl_resource_get_client()` to obtain the `wl_client` from the buffer resource. Given that the buffer resource has just been used, it is guaranteed that the `wl_client` it refers to should be still valid.
    
- In `ViewBackend::dispatchFrameCallbacks()` remove the flush call, and instead do it in `WS::Surface::dispatchFrameCallbacks()`, and pick the `wl_client` from the callback resource objects, if any callback was dispatched at all. Again, if any callback was dispatched their associated `wl_client` should be still valid. Skipping the flush if no callbacks are dispatched is fine because in that case there would not be any messages  pending be sent over the wire anyway to trigger dispatching the callbacks in the client side.

This MR reverts half a dozen commits which were attempting to handle destruction of the `wl_client` gracefully, because after going in circles for a couple of weeks I think that the best approach is the one suggested in the last commit of this PR and letting libwayland take care of things for us.

Also because of the many reverts, I would review at least the last commit separately, because trying to understand the whole diff from `master` is a bit too much 📜 

----

Fixes #145

Note that this includes now also the changes from #154

### DIFF
--- a/src/ws.cpp
+++ b/src/ws.cpp
@@ -28,6 +28,7 @@
 #include "wpe-audio-server-protocol.h"
 #include "wpe-bridge-server-protocol.h"
 #include "wpe-video-plane-display-dmabuf-server-protocol.h"
+#include <algorithm>
 #include <cassert>
 #include <sys/socket.h>
 #include <unistd.h>
@@ -153,6 +154,7 @@ static const struct wl_compositor_interface s_compositorInterface = {
             [](struct wl_resource* resource)
             {
                 auto* surface = static_cast<Surface*>(wl_resource_get_user_data(resource));
+                WS::Instance::singleton().unregisterSurface(surface);
                 delete surface;
             });
     },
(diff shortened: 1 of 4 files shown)


aperezdc:
Now that the changes from #154 are also included as part of this PR, the only issue remaining is that a web view may stop being updated in the following situation: a page was successfully loaded, then a new load is started which results in a [content filter](https://webkitgtk.org/reference/webkit2gtk/stable/webkit2gtk-4.0-WebKitUserContent.html) blocking the top-level resource, and finally navigation goes to another security origin (and the top-level resource is not blocked).

Both @psaavedra and myself reckon that the above corner case is a rare situation, and that is worth it to have this merged it and putting out a 1.9.91 release which we know will have the crashes tackled by this PR gone—after that we will continue working top iron out the remaining issue.

src/ws.cpp
```diff
@@ -538,15 +539,26 @@ void Instance::unregisterViewBackend(uint32_t bridgeId)
...
 void Instance::dispatchFrameCallbacks(uint32_t bridgeId)
 {
     auto it = m_viewBackendMap.find(bridgeId);
     if (it == m_viewBackendMap.end()) {
-        g_error("Instance::dispatchFrameCallbacks(): "
-                "Cannot find surface with bridgeId %" PRIu32 " in view backend map.", bridgeId);
+        g_warning("Instance::dispatchFrameCallbacks(): "
+                  "Cannot find surface with bridgeId %" PRIu32 " in view backend map. Probably the associated surface is gone due to a premature exit in the client side", bridgeId);
+    }
+    else {
```
zdobersek, line 559: This is a bit mis-formatted.

zdobersek (approved):

src/ws.cpp
aperezdc: Thanks, I'll fix it before landing 😄

PR #154, opened 2021-04-28
Add Instance::unregisterSurface() in the destroy funct 

psaavedra:
... of the create surface method of the `struct wl_compositor_interface`

This commits prevents crashes by accessing to already deleted Surfaces destroyed by a premature wl_client_destroy method call. For example, trying to load a non-allowed URI filtered by a Content-Filter policy.

Changes:

* The Instance class implements an Instance::unregisterSurface().

* The destroy callback in the wl_resource_set_implementation for the created Surface now includes a Instance::unregisterSurface(). For this removes the surface from the ViewBackendMap avoiding dispatch operations over a already destroyed Surface.

* The Instance::dispatchFrameCallbacks(uint32_t bridgeId) still checks if the bridgeId has an associated Surface in the ViewBackendMap but now just  logs a warning message and avoids to fail when associated element is not   found. Instead of that the logic now just skips the dispatchFrameCallbacks   if the surface is there.

Co-authored-by: Adrian Perez de Castro <aperez@igalia.com>
Acked-by: Adrian Perez de Castro <aperez@igalia.com>

### DIFF
--- a/src/ws.cpp
+++ b/src/ws.cpp
@@ -28,6 +28,7 @@
 #include "wpe-audio-server-protocol.h"
 #include "wpe-bridge-server-protocol.h"
 #include "wpe-video-plane-display-dmabuf-server-protocol.h"
+#include <algorithm>
 #include <cassert>
 #include <sys/socket.h>
 #include <unistd.h>
@@ -153,6 +154,7 @@ static const struct wl_compositor_interface s_compositorInterface = {
             [](struct wl_resource* resource)
             {
                 auto* surface = static_cast<Surface*>(wl_resource_get_user_data(resource));
+                WS::Instance::singleton().unregisterSurface(surface);
                 delete surface;
             });
     },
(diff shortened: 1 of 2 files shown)


src/ws.cpp
```diff
@@ -537,15 +539,30 @@ void Instance::unregisterViewBackend(uint32_t bridgeId)
...
+{
+    auto it = std::find_if(m_viewBackendMap.begin(), m_viewBackendMap.end(),
+        [surface](const std::pair<uint32_t, Surface*>& value) -> bool {
+            return value.second == surface;
+        });
+    if (it != m_viewBackendMap.end()) {
+        g_message( "~~~~~~ Surface %p removed from m_viewBackendMap ~~~~~~", surface);
+        m_viewBackendMap.erase(it);
+    } else {
+        g_message("~~~~~~ Surface %p not found in m_viewBackendMap ~~~~~~", surface);
```
mcatanzaro, line 552: Don't forget to remove the debug!

psaavedra: Sure. Fixed.

psaavedra:
This patch fix the crash but still I'm observing some issues. In the scenario:

* Load page A
* Load page B (filtered by Content-Policy)
* Load page C

the loading process  of C keeps frozen after a couple of surface frame()/commit() iterations.

aperezdc:
Let's merge this into the `aperezdc/fixfix` branch—I think we need no review for that, and it really makes things much better. There is still one corner case we need to iron out, but at least with this we won't have any more crashes.

aperezdc (approved):

PR #156, opened 2021-04-30
Set back the ViewBackend to the previous bridge references

psaavedra:
Falling back to the previous bridge Id because the current Surface is gone. This action prevents inconsistencies by keeping the ViewBackend in a consistent state pointing to a still valid bridge reference.

An example of a scenario that could generate this kind of inconsistency is a failed load request due to content filter policies. In that case, the client gone situation is reached because

Co-authored-by: Adrian Perez de Castro <aperez@igalia.com>
Acked-by: Adrian Perez de Castro <aperez@igalia.com>


psaavedra:
Replaced by https://github.com/Igalia/WPEBackend-fdo/pull/159

PR #157, opened 2021-04-30
Partially revert a4eb79d, video-plane and audio can be NULL

aperezdc:
Remove wrong `NULL` checks for variables which are actually allowed to be `NULL` during normal operation.

The global objects for the `wpe_audio` and `wpe_video_plane_display_dmabuf` interfaces are created dynamically when a receiver is registered for them, which means at the time the nested compositor client is connected they may be `NULL`.

### DIFF
--- a/src/extensions/audio.cpp
+++ b/src/extensions/audio.cpp
@@ -159,9 +159,6 @@ public:
         wl_display_roundtrip_queue(display, eventQueue);
         wl_registry_destroy(registry);
 
-        if (!m_wl.audio)
-            g_error("Failed to bind wpe_audio");
-
         wl_event_queue_destroy(eventQueue);
     }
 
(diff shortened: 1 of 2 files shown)


mcatanzaro (approved):
Oops, you are correct, it doesn't make sense to do g_error() when there are NULL checks in the rest of the code.

Note the Audio and DmaBuf objects will not actually *do anything* because there is no way for `m_wl.audio` and `m_wl.videoPlaneDisplayDmaBuf` to ever be set later on, but I guess that's intended.

aperezdc:
> Note the Audio and DmaBuf objects will not actually _do anything_ because there is no way for `m_wl.audio` and `m_wl.videoPlaneDisplayDmaBuf` to ever be set later on, but I guess that's intended.

Yes, my understanding is also that this is intended behaviour. @philn would be the one to know with 100% certainty 😄

aperezdc:
Cherry-picked into the `wpebackend-fdo-1.10` branch as commit 8a021ec9d6f11cb6790841469bf826c9c1a9e2e1

PR #158, opened 2021-05-04
Keep list of bridgeIds associated with each ViewBackend

### LINKED ISSUE #155: Navigation after top-level resource blocked sometimes results in unresponsive web view
[…] the only issue remaining is that a web view may stop being updated in the following situation: a page was successfully loaded, then a new load is started which results in a [content filter](https://webkitgtk.org/reference/webkit2gtk/stable/webkit2gtk-4.0-WebKitUserContent.html) blocking the top-level resource, and finally navigation goes to another security origin (and the top-level resource is not blocked).

Both @psaavedra and myself reckon that the above corner case is a rare situation, and that is worth it to have this merged it and putting out a 1.9.91 release which we know will have the crashes tackled by this PR gone—after that we will continue working top iron out the remaining issue.

_Originally posted by @aperezdc in https://github.com/Igalia/WPEBackend-fdo/issues/153#issuecomment-830045466_

aperezdc:
As each ViewBackend (and hence web view) can get its contents rendered by different WPEWebProcess due to PSON, keep around a list of all the differen bridgeIds which have been associated with it in a `std::vector`. Newly added elements are always pushed at the back, so the most recent association (i.e. the “current” active one) is always the last element. Keeping the list allows getting back to the previously used ones.

----

Fixes #155

Note that this may need some testing with the `devel/multiple-views` and under WebKitGTK before merging.

This is an alternative to #156 which implements a general solution.

### DIFF
--- a/src/view-backend-private.cpp
+++ b/src/view-backend-private.cpp
@@ -28,6 +28,7 @@
 #include <cassert>
 #include <sys/types.h>
 #include <sys/socket.h>
+#include <algorithm>
 
 ViewBackend::ViewBackend(ClientBundle* clientBundle, struct wpe_view_backend* backend)
     : m_clientBundle(clientBundle)
@@ -38,7 +39,8 @@ ViewBackend::ViewBackend(ClientBundle* clientBundle, struct wpe_view_backend* ba
 
 ViewBackend::~ViewBackend()
 {
-    unregisterSurface(m_bridgeId);
+    while (!m_bridgeIds.empty())
+        unregisterSurface(m_bridgeIds.front());
 
     if (m_clientFd != -1)
         close(m_clientFd);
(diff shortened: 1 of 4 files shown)


aperezdc:
I have added one more commit with a slightly modified version of @psaavedra's very good suggestion from #159; this should be good to go now 😄

carlosgcampos (approved):

src/view-backend-private.cpp
```diff
@@ -92,10 +94,10 @@ void ViewBackend::exportEGLStreamProducer(struct wl_resource* bufferResource)
 
 void ViewBackend::dispatchFrameCallbacks()
 {
-    if (G_LIKELY(m_bridgeId))
-        WS::Instance::singleton().dispatchFrameCallbacks(m_bridgeId);
-
-    wpe_view_backend_dispatch_frame_displayed(m_backend);
+    if (G_LIKELY(!m_bridgeIds.empty())) {
```
mcatanzaro, line 97: Is it really likely, though? Yeah, there's *probably* been a process swap, but if not it will tank performance, right? I don't like using G_LIKELY() unless I'm confident the case is extremely rare.

aperezdc: The most common case is that there is at least one `bridgeId` that associates the surface with a `ViewBackend`; this code is hit for every rendered frame and that is my reasoning for using `G_LIKELY()` here.

aperezdc:
Given that we have had this approved for some days now, that it has received a week of testing, and that it improves reliability when using multiple views (as it avoids most cases of views not being updated), today in a chat with @psaavedra we have agreed to merge this PR so I can make a 1.9.92 release and include it as a backport in a 1.8.4 release as well.

PR #160, opened 2021-05-05
meson: Try wayland-scanner first without pkg-config

aperezdc:
Try finding the wayland-scanner program first without involving picking its path from the `pkg-config` dependency. This way it is possible to instruct Meson to override the the path where to find it using the `[binaries]` section of a [native file](https://mesonbuild.com/Machine-files.html) during cross-compilation.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -91,15 +91,13 @@ add_project_arguments(
 	language: ['c', 'cpp']
 )
 
+wayland_scanner_candidates = ['wayland-scanner']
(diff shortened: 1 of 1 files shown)


carlosgcampos (approved):

aperezdc:
Cherry-picked into the `wpebackend-fdo-1.10` branch as commit b86b0db2f8d6a784357051acfee30d4b9ef7c10c

PR #164, opened 2021-06-08
meson: Add cpp_rtti=false default option

aperezdc:
Add the missing `cpp_rtti=false` option in the list of defaults, which will be picked by Meson 0.53 or newer. Older versions will keep using the existing check.


bertogg (approved):
Tested, it works fine

aperezdc:
Thanks @bertogg!

aperezdc:
Cherry-picked into the `wpebackend-1.10` branch as commit a1dcb01ce5763806fde709cbe01cbd604fa6693b

PR #165, opened 2021-06-08
CI: Add task to run the Clang static analyzer

aperezdc:
Run the Clang static analyzer as a CI job. Instead of using `scan-build`, which uses `LD_PRELOAD` to interpose the compiler, install the Python scan-build package which includes an `analyze-build` program that can directly ingest the JSON compilation database.


zdobersek (approved):

nikolaszimmermann:
Nice one!

PR #167, opened 2021-09-02
Meson updates

ceyusa:
meson: Use implicit dependencies for required dependencies.
    
    The generated pkgconfig file for the library is incomplete, since application,
    such as cog, only explicity depend on WPEBackend-fdo, though it also depends on
    libwpe. It has not been a problem since wpe-webkit-1.0.pc explicity expose
    libwpe requirement. Still, for robustness and self documentation, it's better to
    express that dependency explicity too.
    
    Also, this patch removes the pkg config version declation since it's also
    handled implicitly by meson.
    
    Previous pc file:
    
    prefix=/usr
    libdir=${prefix}/lib/x86_64-linux-gnu
    includedir=${prefix}/include
    
    Name: wpebackend-fdo-1.0
    Description: The WPEBackend-fdo library
    Version: 1.9.91
    Libs: -L${libdir} -lWPEBackend-fdo-1.0
    Cflags: -I${includedir}/wpe-fdo-1.0
    
    New pc file:
    
    prefix=/usr/local
    libdir=${prefix}/lib
    includedir=${prefix}/include
    
    Name: wpebackend-fdo-1.0
    Description: The WPEBackend-fdo library
    Version: 1.9.1
    Requires: wpe-1.0
    Requires.private: xkbcommon, egl
    Libs: -L${libdir} -lWPEBackend-fdo-1.0
    Libs.private: -ldl
    Cflags: -I${includedir}/wpe-fdo-1.0

----

    meson: Support meson's subproject layout.
    
    As libwpe, wpebackend-fdo can be used as a meson's subproject.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -254,13 +254,15 @@ add_project_arguments(
 	language: 'c'
 )
 
+wpe_dep = dependency('wpe-1.0', version: '>=1.6.0', fallback: ['libwpe', 'libwpe_dep'])
+
 deps = [
 	dependency('epoxy'),
 	dependency('gio-2.0'),
 	dependency('gobject-2.0'),
 	dependency('wayland-client'),
 	dependency('wayland-server'),
-	dependency('wpe-1.0', version: '>=1.6.0', fallback: ['libwpe', 'libwpe_dep']),
+        wpe_dep,
 ]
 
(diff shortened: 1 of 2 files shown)


aperezdc (approved):
Thanks for the improvements, @ceyusa! 👍🏼

PR #176, opened 2022-07-09
Fix SIGSEGV in WebKit2 applications

vieiraa:
Detected this SIGSEGV when trying to login to game stores in Lutris.
May fix #175 
Also fixes crashes in Epiphany, and possibly in other WebKit2 applications as well.

Issue #175, opened 2022-07-08
Invalid write of size 1 in releaseImage


jameshilliard:
I'm seeing this get flagged by valgrind

```cpp
==17== Invalid write of size 1
==17==    at 0x19AED72B: releaseImage (view-backend-exportable-fdo-egl.cpp:250)
==17==    by 0x19AED72B: wpe_view_backend_exportable_fdo_egl_dispatch_release_exported_image (view-backend-exportable-fdo-egl.cpp:334)
==17==    by 0x1B5D769E: render (in /app/webkit/WebKitBuild/Debug/Tools/cog-prefix/src/cog-build/modules/libcogplatform-gtk4.so)
==17==    by 0x1B69DB90: _gtk_marshal_BOOLEAN__OBJECTv (gtkmarshalers.c:876)
==17==    by 0x15F18228: _g_closure_invoke_va (gclosure.c:893)
==17==    by 0x15F30E4A: g_signal_emit_valist (gsignal.c:3406)
==17==    by 0x15F31D62: g_signal_emit (gsignal.c:3553)
==17==    by 0x1B7865F8: gtk_gl_area_snapshot (gtkglarea.c:729)
==17==    by 0x1B8F38BE: gtk_widget_create_render_node (gtkwidget.c:11587)
==17==    by 0x1B8F6783: gtk_widget_do_snapshot (gtkwidget.c:11622)
==17==    by 0x1B903452: gtk_widget_snapshot_child (gtkwidget.c:12043)
==17==    by 0x1B9034DD: gtk_widget_real_snapshot (gtkwidget.c:734)
==17==    by 0x1B8F38BE: gtk_widget_create_render_node (gtkwidget.c:11587)
==17==    by 0x1B8F6783: gtk_widget_do_snapshot (gtkwidget.c:11622)
==17==    by 0x1B903452: gtk_widget_snapshot_child (gtkwidget.c:12043)
==17==    by 0x1B9034DD: gtk_widget_real_snapshot (gtkwidget.c:734)
==17==    by 0x1B8F3C95: gtk_widget_create_render_node (gtkwidget.c:11582)
==17==    by 0x1B8F6783: gtk_widget_do_snapshot (gtkwidget.c:11622)
==17==    by 0x1B902C72: gtk_widget_snapshot (gtkwidget.c:11644)
==17==    by 0x1B902C72: gtk_widget_render (gtkwidget.c:11676)
[...]

PR #178, opened 2022-12-01
Remove image on buffer destroy callback only

psaavedra:
view-backend-exportable-fdo-egl.cpp relies on the bufferDestroyListenerCallback for destroying the image, This is called during the wl_resource_destroy() to release the image.

Related-to: #175 #176

### DIFF
--- a/src/view-backend-exportable-fdo-egl-private.h
+++ b/src/view-backend-exportable-fdo-egl-private.h
@@ -33,7 +33,6 @@ struct wpe_fdo_egl_exported_image {
     EGLImageKHR eglImage { nullptr };
     uint32_t width { 0 };
     uint32_t height { 0 };
-    bool exported { false };
     struct wl_resource* bufferResource { nullptr };
     struct wl_listener bufferDestroyListener;
 };
--- a/src/view-backend-exportable-fdo-egl.cpp
+++ b/src/view-backend-exportable-fdo-egl.cpp
@@ -249,8 +249,6 @@ public:
     {
         if (image->bufferResource)
             viewBackend->releaseBuffer(image->bufferResource);
-        else
-            deleteImage(image);
     }
 
     void releaseShmBuffer(struct wpe_fdo_shm_exported_buffer* buffer)
@@ -277,7 +275,6 @@ private:
 
     void exportImage(struct wpe_fdo_egl_exported_image* image)
     {
-        image->exported = true;
         client->export_fdo_egl_image(data, image);
     }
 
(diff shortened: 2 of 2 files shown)


psaavedra:
Valgrind's traces on #175 

```
==17==  Address 0x1fac1fd0 is 0 bytes inside a block of size 56 free'd
==17==    at 0x48438AF: operator delete(void*) (vg_replace_malloc.c:923)
==17==    by 0x1A8674F5: wl_priv_signal_final_emit (wayland-server.c:2221)
==17==    by 0x1A8674F5: destroy_resource (wayland-server.c:724)
==17==    by 0x1A867C24: wl_resource_destroy (wayland-server.c:744)
==17==    by 0x1A099FE4: ffi_call_unix64 (unix64.S:101)
==17==    by 0x1A0993F5: ffi_call_int (ffi64.c:669)
==17==    by 0x1A86D2A1: wl_closure_invoke (connection.c:1025)
==17==    by 0x1A868216: wl_client_connection_data (wayland-server.c:437)
==17==    by 0x1A86B019: wl_event_loop_dispatch (event-loop.c:1027)
```


shows that the first deletion happens during the destruction of the wl_client because a problem in the `wl_client_connection_data()`.

On

```c
static int
wl_client_connection_data(int fd, uint32_t mask, void *data)
{
...
        if ((resource_flags & WL_MAP_ENTRY_LEGACY) ||
            resource->dispatcher == NULL) {
            wl_closure_invoke(closure, WL_CLOSURE_INVOKE_SERVER,
                      object, opcode, client);
```

this part of the code is reached if  `resource->dispatcher == NULL)` so probably this is 
 because the peer on the channel is already gone at than point and wayland decides to call `wl_closure_invoke()`.

aperezdc:
I can take a look at this later today, but @carlosgcampos and/or @zdobersek may want to comment here as well.

alexgcastro:
We are already reviewing the patch, hopefully we will have a resolution this week. Without the option to reproduce the original problems it took some time to check the lifecycle of the object considering that image structure position between wayland and the backend.

alexgcastro (approved):
The code apparently missed the situation where wayland destruction and the releaseImage happened at the same time. The places in the code that are checking if bufferResource is NULL are doing the work but was not enough, that explains previous patches. Still the code needs a better name and check for that situation, but this patch is in the right direction. We have to handle the image in the desctruction and avoid any operation when the bufferResource is gone. The option implemeted was keep the image and hope for the best.

It is a pity we can not reproduce the original valgrind stacks and crashes to verify the whole situation though.

Thanks for the patch, hopefully we can refactor this code at some point.

aperezdc:
Thanks a lot @alexgcastro! 🤗

aperezdc:
Cherry-picked into the `wpebackend-fdo-1.14` branch as commit 0d6a75a61e8377d65130eeb59b752cac8e9c9fca.

PR #179, opened 2022-12-07
Add Bug report issue template

psaavedra:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,43 @@
...
+    1. The build command line causing the error
+    1. The error output
+1. In case of a runtime error
+    1. The execution command line causing the error
+    1. Description of  the error
+    1. Backtrace of the coredump (in case of)
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
```
mariospr, line 33: Can we add an "**Actual behavior**" section here too? I suggest this:

```
**Actual behavior**
A clear and concise description of what it actually happened.
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,43 @@
...
+assignees: ''
+
+---
+
+**Bug description**
+A clear and concise description of what the bug is.
+
+**How To Reproduce**
+
+Fikes to reproduce the behavior:
```
mariospr, line 15: I'm not sure what "fikes" mean in this context... maybe it's a typo and should be "Files" instead?

I think this should be `Relevant information and input files to reproduce the behavior`, which I think it better matches what we are requesting here.

On top of that change, a nit: I'd **also change the list into a non-numbered bullet item list** (instead of ordered), since there's no need to provide that information in any particular order.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,43 @@
...
+labels: bug
+assignees: ''
+
+---
+
+**Bug description**
+A clear and concise description of what the bug is.
+
+**How To Reproduce**
+
```
mariospr, line 14: I think both input files/data **and** steps are important to reproduce an issue, so I'd suggest to add another list before the one below titled `Steps to reproduce the behavior`, something like this:
```
Steps to reproduce the behavior
  * Step 1
  * Step 2
  * ...
```

mariospr (approved):
To me this looks good, but I have a couple of suggestions that I think would make sense to apply **before merging**.

mariospr (changes requested):
(clicked on the wrong item before, sorry)

manuelafm (approved):
LGTM

mariospr (approved):
LGTM

PR #180, opened 2022-12-19
Fix assert for NVidia drivers 525.60.11

neodesys:
With this specific version of NVidia drivers, modifiers cannot be queried for specific formats, so instead of raising an assertion let's just ignore the format silently.

### DIFF
--- a/src/ws-egl.cpp
+++ b/src/ws-egl.cpp
@@ -294,7 +294,7 @@ void ImplEGL::foreachDmaBufModifier(std::function<void (int format, uint64_t mod
         uint64_t modifiers[64];
         EGLint numModifiers;
         if (!s_eglQueryDmaBufModifiersEXT(m_egl.display, formats[i], 64, modifiers, NULL, &numModifiers))
-            assert(!"Linux-dmabuf: Failed to query modifiers of a format");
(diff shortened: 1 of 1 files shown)


philn:
@zdobersek do you think you can review this?

src/ws-egl.cpp
```diff
@@ -294,7 +294,7 @@ void ImplEGL::foreachDmaBufModifier(std::function<void (int format, uint64_t mod
         uint64_t modifiers[64];
         EGLint numModifiers;
         if (!s_eglQueryDmaBufModifiersEXT(m_egl.display, formats[i], 64, modifiers, NULL, &numModifiers))
-            assert(!"Linux-dmabuf: Failed to query modifiers of a format");
+            continue;
```
zdobersek, line 297: When this fails, `numModifiers` should be set to 0, allowing `DRM_FORMAT_MOD_INVALID` to be passed through. This way the format doesn't go ignored.

zdobersek (changes requested):

src/ws-egl.cpp
neodesys: Done :+1:

zdobersek (approved):

PR #181, opened 2023-02-15
view-backend: dispatch pending frame callbacks after a surface is unr…

carlosgcampos:
…egistered

This was causing WebKit to stop painting after provisional navigation fails with PSON enabled. It happened in the following situation:

1. Normal load, view backend registers surface for bridge 1
2. A frame is requested for bridge 1
3. CORS load starts, a new process is launched and view backend registers surface for bridge 2
4. Frame for bridge 1 is rendered, but current bridge is 2 so ViewBackend::dispatchFrameCallbacks() does nothing.
5. CORS load fails and the provisional load is destroyed so surface for bridge 2 is unregistered
6. No more updates because there's still the frame callback pending for surface for bridge 1.

This always happens when WebKit starts a CORS navigation that is converted to a download, because in that case the provisional load fails with cancelled error. It was never committed and nothing was rendered. So, while it's ok to stop painting new frames during the provisional load to avoid flashing, we should check if there's any pending callback once the current surface is unregistered.

### DIFF
--- a/src/view-backend-private.cpp
+++ b/src/view-backend-private.cpp
@@ -105,8 +105,8 @@ void ViewBackend::commitDmabufPoolEntry(struct wpe_dmabuf_pool_entry* entry)
 void ViewBackend::dispatchFrameCallbacks()
 {
     if (G_LIKELY(!m_bridgeIds.empty())) {
-        WS::Instance::singleton().dispatchFrameCallbacks(m_bridgeIds.back());
-        wpe_view_backend_dispatch_frame_displayed(m_backend);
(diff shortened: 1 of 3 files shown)


aperezdc (approved):
Patch LGTM, and the description of the issue in the commit log helped quite a bit to understand both the issue and the solution. You may want to have another pair of eyes check this one (maybe @zdobersek?) but I think we should merge a fix for this sooner better than later.

Also, let's backport it to the current stable series 🔙

aperezdc:
Cherry-picked into the `wpebackend-fdo-1.14` branch as commit 3185594b65170144fbc35df13ad5f0b83bc58450.

PR #184, opened 2023-11-23
ws: Save and expose damage buffer information

lauromoura:
The wl_surface.damage_buffer request can be used by clients to inform that a given region (in buffer coordinates) is damaged. This request is used by eglSwapBuffersWithDamage[EXT|KHR], with one call for each rect.

For more info: https://wayland-book.com/surfaces-in-depth/damaging-surfaces.html

This commit stores this information and exposes it to exported EGL image clients.

It also bumps the nested compositor interface to version 4, which provides the damage_buffer support.


lauromoura:
Superseded by #188

PR #187, opened 2024-03-13
doc: Document EGL exported images

aperezdc:
...because some documentation is always better than no documentation.


TingPing (approved):
I believe the default for the arguments is `(transfer none)`, but seems fine.

aperezdc:
> I believe the default for the arguments is `(transfer none)`, but seems fine.

Double checked, and yes you are right. I always forget what's the default so I lean towards being explicit which does no harm 😉 

Thanks!

PR #188, opened 2024-03-16
ws: Save and expose damaged image regions (v2)

aperezdc:
This builds upon #184, with the following additions:

- Meson will check that the Wayland headers define the needed symbols to use `wl_surface_damage_buffer()`, erroring if not available. The functionality is now always built-in. This should not be an issue because any reasonable Wayland version should have it.
- Plugs a leak that would leave stale entries in the `unordered_map` that associates a buffer with its damaged regions list when a buffer is destroyed.
- Stores damage regions using a new `wpe_fdo_rect` type, which makes the code easier to follow, and documents the damage support, explicitly indicating that top-left is used as origin of coordinates for damage.
- Adds warnings to make developers notice if any of the stubbed methods from `wl_surface_interface` ends up being used.

Thanks to @lauromoura for the first version of the patch set 🙏🏼


aperezdc:
We won't be merging this; we have a similar mechanism in WPEPlatform, which is preferred.

PR #190, opened 2024-06-25
REGRESSION(b51f539): Fix memory leak when the buffer is destroyed at view-backend-exportable-fdo-egl

clopez:
Since commit b51f539 there is a memory leak each time the wl_resource is destroyed. This can be easily reproduced by repeteadly switching full-screen on/off (pressing F11 key) with Cog on Weston.

The memory leak is caused because since b51f539 the wpe_fdo_egl_exported_image object is not cleaned anymore on the bufferDestroyListenerCallback callback.

Commit cb6b86a fixed the leak but introduced crashes on some cases, so it was reverted.

This is a new attempt at fixing this leak, this adds safeguards to ensure that the image object is not cleaned twice or with the wrong exported status.

Related-to: #73 #175 #176 #178
Related-to: https://github.com/Igalia/cog/issues/538

### DIFF
--- a/src/view-backend-exportable-fdo-egl.cpp
+++ b/src/view-backend-exportable-fdo-egl.cpp
@@ -247,9 +247,13 @@ public:
 
     void releaseImage(struct wpe_fdo_egl_exported_image* image)
     {
-        if (image->bufferResource)
-            viewBackend->releaseBuffer(image->bufferResource);
-        else
+        if (!image)
+            return;
+        if (image->exported) {
+            image->exported = false;
+            if (image->bufferResource)
+                viewBackend->releaseBuffer(image->bufferResource);
+        } else
             deleteImage(image);
     }
 
@@ -285,7 +289,6 @@ private:
     {
         assert(image->eglImage);
         WS::instanceImpl<WS::ImplEGL>().destroyImage(image->eglImage);
-
         delete image;
     }
 
@@ -293,8 +296,13 @@ private:
     {
         struct wpe_fdo_egl_exported_image* image;
         image = wl_container_of(listener, image, bufferDestroyListener);
-
-        image->bufferResource = nullptr;
+        // The image object is inmediately destroyed here only if it's not currently
+        // exported, otherwise it's destroyed when the client releases/returns it.
+        if (image->exported) {
+            image->exported = false;
+            image->bufferResource = nullptr;
+        } else
+            deleteImage(image);
     }
 };
 


psaavedra (commented):
I've tested this change in 2.44. I can say it looks good to me so far. No crashes nor memory leaks. 


Still I want to read and understand more in deep the implications of the change.

src/view-backend-exportable-fdo-egl.cpp
```diff
@@ -247,9 +247,13 @@ class ClientBundleEGL final : public ClientBundle {
 
     void releaseImage(struct wpe_fdo_egl_exported_image* image)
     {
-        if (image->bufferResource)
-            viewBackend->releaseBuffer(image->bufferResource);
-        else
+        if (!image)
+            return;
+        if (image->exported) {
+            image->exported = false;
```
psaavedra, line 253: I understand this change restores the first logic (`exported` safe-guard ) implemented in the original commit https://github.com/Igalia/WPEBackend-fdo/commit/fb897d2972a0bbc4d0157512f3da281ec7be917f#diff-a2ff2fb4cb225868a0a9b4ae69769ab3efc9e6fbb06d26885c123582bc7a3a0bR198 which it was not problematic at all. This looks good.

src/view-backend-exportable-fdo-egl.cpp
```diff
@@ -285,16 +289,20 @@ class ClientBundleEGL final : public ClientBundle {
...
     static void bufferDestroyListenerCallback(struct wl_listener* listener, void*)
     {
         struct wpe_fdo_egl_exported_image* image;
         image = wl_container_of(listener, image, bufferDestroyListener);
-
-        image->bufferResource = nullptr;
+        // The image object is inmediately destroyed here only if it's not currently
+        // exported, otherwise it's destroyed when the client releases/returns it.
+        if (image->exported) {
+            image->exported = false;
```
psaavedra, line 302: Good. This forces entering in the  `deleteImage(image)` after in the `releaseImage(struct wpe_fdo_egl_exported_image* image)`.

psaavedra (approved):
LGTM since this eventually reverts the state of the code to something similar to https://github.com/Igalia/WPEBackend-fdo/commit/fb897d2972a0bbc4d0157512f3da281ec7be917f#diff-a2ff2fb4cb225868a0a9b4ae69769ab3efc9e6fbb06d26885c123582bc7a3a0b what is not problematic.

PR #193, opened 2024-08-28
Fix wpe_dmabuf_pool object leak

rawoul:
Fixes wayland library warning on exit:
```
warning: queue 0x887860 destroyed while proxies still attached:
  wpe_dmabuf_pool@12 still attached
```


aperezdc (approved):
Thanks @rawoul, good catch!

PR #199, opened 2025-03-17
CI: Update to newer Ubuntu, Actions, and simplify

aperezdc:
Update the CI images used to Ubuntu 22.04, which allows removing the usage of actions/setup-python, now that the base system includes a newer Python version.

Update to actions/cache and actions/checkout v4.

Improve usage of actions/cache, by also caching virtualenvs used for building.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,31 +6,46 @@ on: [push]
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Install Debian Packages
         run: |
           sudo apt update
-          sudo apt install -y flex libjson-glib-dev libxkbcommon-dev \
-            libegl1-mesa-dev libxml2-dev libxslt1-dev libyaml-dev \
-            libwayland-dev llvm-dev libclang-dev libglib2.0-dev \
-            libepoxy-dev ninja-build
-      - name: Setup Python
-        uses: actions/setup-python@v4
-        with:
-          python-version: '3.6'
+          sudo apt install -y \
+            cmake \
+            flex \
+            libclang-dev \
+            libegl1-mesa-dev \
+            libepoxy-dev \
+            libglib2.0-dev \
+            libjson-glib-dev \
+            libwayland-dev \
+            libxkbcommon-dev \
+            libxml2-dev \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -6,31 +6,46 @@ on: [push]
...
       - name: Install Python Packages
+        if: steps.cache.outputs.cache-hit != 'true'
+        env:
+          HOTDOC_BUILD_C_EXTENSION: enabled
         run: |
-          python -m pip install --upgrade pip setuptools wheel
-          HOTDOC_BUILD_C_EXTENSION=enabled pip install hotdoc meson==0.55
+          python3 -m venv --upgrade-deps .venv
+          . .venv/bin/activate
+          pip install meson==0.55 hotdoc scan-build
```
dpino, line 48: It seems Ubuntu 22.04 features meson 0.61.

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.5 LTS
Release:        22.04
Codename:       jammy
```
```bash
$ apt-cache madison meson
     meson |   0.61.2-1 | http://ports.ubuntu.com/ubuntu-ports jammy/universe arm64 Packages
```

So unless you need meson 0.55, it would be simpler to install the `meson` package that comes with the distribution.

dpino (approved):

.github/workflows/ci.yml
aperezdc: We want to keep testing with the oldest Meson we are supposed to support. As a matter of fact, we should apply this change to `meson.build` in a separate commit:

```diff
diff --git a/meson.build b/meson.build
index 634aa46..420cf72 100644
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,5 @@
 project('wpebackend-fdo', ['c', 'cpp'],
-	meson_version: '>=0.52.1',
+	meson_version: '>=0.55',
 	default_options: [
 		'b_ndebug=if-release',
 		'c_std=c99',
```

😉

aperezdc:
Cherry-picked (with amendments) into `wpebackend-fdo-1.16`, commits 7c14d3e43c7a514a27f65e0e2110cee3087d47bb, 491ee4ef9bb5ea022319923744c923405c0ff337, and 33863c3ce25120515d81730916792bf5736f9c58

aperezdc:
I have skipped the backport to `wpebackend-1.14` as that's the old stable branch, and not expected to receive further updates.

